PR #41, opened 2014-12-21
Don't set a default whole request timeout.

utkarshkukreti:
I don't think a binding should set such a low value (or any value) by default.

http://curl.haxx.se/libcurl/c/CURLOPT_TIMEOUT_MS.html

### DIFF
--- a/src/http/handle.rs
+++ b/src/http/handle.rs
@@ -21,7 +21,6 @@ pub struct Handle {
 impl Handle {
     pub fn new() -> Handle {
         return configure(Handle { easy: Easy::new() }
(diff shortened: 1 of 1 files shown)


carllerche:
Pretty late, but I don't personally have a huge opinion. I don't like having unlimited timeouts. I think @alexcrichton likes having a 30 second default timeout.

alexcrichton:
Hey @utkarshkukreti, I'm taking over the curl crate for now and I've got an API rewrite over at https://github.com/alexcrichton/curl-rust/pull/105 which I believe should address this as well. If you'd like to take a look though feel free!

PR #52, opened 2015-01-05
Update to rustc 0.13.0-nightly (ad9e75938 2015-01-05 00:26:28 +0000).

saneyuki:
related: https://github.com/rust-lang/rust/commit/400c3a0d


saneyuki:
duplicated: #51

PR #53, opened 2015-01-07
(Hopefully) fixes changes in FFI

globin:
Please be careful as the proxy test is failing and I have no idea if
that is caused by these changes. At least it should be a starting point
to continue on from.


alexcrichton:
Thanks!

PR #55, opened 2015-01-10
update to rustc 1.0.0-dev (73a25f55a 2015-01-09 20:10:07 +0000)

gifnksm:


alexcrichton:
Argh lost in the rebase when merging #56, but thanks regardless!

PR #64, opened 2015-03-20
fix broken formatting of resp body and headers in the example of readme....

samuelleeuwenburg:
...md

Was not able to use the example without this fix. Might be useful to update for new users.


carllerche:
Thanks! Merged: 0090605199e809956a6c97922b04be899637c658

PR #67, opened 2015-04-10
Added a convenience method for making PATCH requests

davidrhyswhite:
Added a convenience method for making PATCH requests:

```rust
    let srv = server!(
        recv!(b"PATCH / HTTP/1.1\r\n\
                Host: localhost:{PORT}\r\n\
                Accept: */*\r\n\
                Content-Type: application/octet-stream\r\n\
                Content-Length: 11\r\n\
                \r\n\
                Foo Bar Baz"),
        send!(b"HTTP/1.1 200 OK\r\n\
                Content-Length: 5\r\n\r\n\
                Hello\r\n\r\n")
    );

    let res = http::handle()
        .patch(server::url("/"), "Foo Bar Baz")
        .exec().unwrap();
```


carllerche:
Looks good to me. cc @alexcrichton

PR #74, opened 2015-06-11
An example of curl_multi

posix4e:
I'm looking at adding support for the curl_multi api. This creates a lot of duplication (which I have tried to avoid) and most likely we will need a different library that pulls in eventual and mio to actually make use of the bindings. I'm working on one called micro-hyper. Consider this pull a demonstration and a WIP. The question I have is what the api should look like.

### DIFF
--- a/curl-sys/lib.rs
+++ b/curl-sys/lib.rs
@@ -8,8 +8,11 @@ use libc::{c_void, c_int, c_char, c_uint, c_long};
 
 pub type CURLINFO = c_int;
 pub type CURL = c_void;
+pub type CURLM = c_void;
 pub type curl_slist = c_void;
 pub type CURLoption = c_int;
+pub type CURLMoption = c_int;
+pub type C = c_int;
 
 #[repr(C)]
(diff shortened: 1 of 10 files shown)


carllerche:
Sorry for the delay. Trying to catch up.

I definitely want to get curl multi bound. Glancing through the PR though, it looks like there is a lot of duplication. Any way to try to avoid duplicating definitions?

posix4e:
No worries. Notice CurlM vs Curl for a lot of the options. What dups are you most worried about?

carllerche:
It's a bit hard for me to follow (and I have lost a lot of context), but I'm noticing some duplication like `CURL_VERSION_IPV6`.

posix4e:
So i removed the easy.rs file altogether.  Is this any clearer?

alexcrichton:
Ok, I've started binding the multi interface in https://github.com/alexcrichton/curl-rust/pull/109, so I'm gonna close this now. Sorry for the delayed response :)

PR #78, opened 2015-07-12
Fix building on OpenBSD

stevenjm:
OpenBSD is very similar to bitrig, so I've just copied what's done for
bitrig and changed the arch triplet. I'm able to build on OpenBSD
-current (as of 2015-07-11) with this patch.


stevenjm:
This is a duplicate of https://github.com/carllerche/curl-rust/pull/72; the diffs are identical. Sorry for the noise.

PR #80, opened 2015-07-31
Added `userpwd` to set curl option `USERPWD`

coeuvre:


carllerche:
Looks like this happened at some point: https://github.com/carllerche/curl-rust/blob/master/src/http/handle.rs#L87-L90

Sorry about the delay!

PR #83, opened 2015-09-01
pass host to curl configuration command during a cross-compilation

Frederick888:


curl-sys/build.rs
```diff
@@ -60,6 +60,15 @@ fn main() {
...
     let mut config_opts = Vec::new();
+    if target.len() > 0 {
+        if windows && target.contains("x86_64") {
+            config_opts.push("--host x86_64-w64-mingw32".to_string());
+        } else if windows && target.contains("i686") {
+            config_opts.push("--host i686-w64-mingw32".to_string());
+        } else {
+            config_opts.push("--host ".to_string() + &target);
+        }
+    }
```
alexcrichton: I think that this can also be done with:

```rust
config_opts.push(format!("--host {}", target.replace("pc-windows-gnu", "w64-mingw32")));
```

Frederick888: @alexcrichton Oh, right, it's a nice simplification.

curl-sys/build.rs
```diff
@@ -60,6 +60,9 @@ fn main() {
     let _ = fs::create_dir(&dst.join("build"));
 
     let mut config_opts = Vec::new();
+    if target.len() > 0 {
```
alexcrichton: I think you can also omit this check because this should always be true I believe.

alexcrichton:
Hey @Frederick888, I'm taking over the curl crate for now and I've got an API rewrite over at https://github.com/alexcrichton/curl-rust/pull/105 which I believe should address this as well. If you'd like to take a look though feel free!

PR #85, opened 2015-09-13
Remove patch level from versions

mneumann:


alexcrichton:
I've got a rewrite over at https://github.com/alexcrichton/curl-rust/pull/105 which I think should obviate the need for this, but feel free to take a look!

PR #86, opened 2015-09-24
Allow to use deprecated `connect()` for Rust 1.1 support

zetok:


alexcrichton:
Thanks for the PR! Currently this library is still targeting rust 1.1, so these methods can't be renamed just yet, but perhaps an `#[allow(deprecated)]` annotation could be added?

zetok:
Well, I guess I'll morph this PR to do that.
(not ~now though, I'll do that probably within a day)

zetok:
ok, done.

alexcrichton:
Thanks!

PR #88, opened 2015-10-14
Android support in Cargo.toml.

overminder:
Previously the openssl dependencies were not declared for Android targets.

Also changed the cargo build scripts a bit to mention that we are trying to
compile libcurl from source. This can be helpful if the compilation fails
and the user wants to know why.

(Compiling libcurl from source didn't fail for me - I just wanted to use my prebuilt cross-compiled libcurl by setting PKG_CONFIG_PATH but I didn't know PKG_CONFIG_ALLOW_CROSS before..)

### DIFF
--- a/curl-sys/build.rs
+++ b/curl-sys/build.rs
@@ -4,6 +4,7 @@ extern crate gcc;
 use std::env;
 use std::ffi::OsString;
 use std::fs;
+use std::io::{stderr, Write};
 use std::path::PathBuf;
 use std::process::Command;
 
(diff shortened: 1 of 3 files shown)


curl-sys/build.rs
```diff
@@ -30,7 +31,8 @@ fn main() {
     // Next, fall back and try to use pkg-config if its available.
     match pkg_config::find_library("libcurl") {
         Ok(..) => return,
-        Err(..) => {}
+        Err(e) => writeln!(&mut stderr(), "Couldn't find libcurl from \
+                           pkgconfig ({:?}), compiling it from source...", e).unwrap(),
```
alexcrichton: Could this just be a println?

overminder: I used stderr since I thought stdout in the build script is strictly for cargo configuration and printing arbitrary lines to stdout might confuse cargo. So that's not the case?

alexcrichton: Ah nah it's fine to dump all other kinds of data to stdout so long as it doesn't look like "cargo:foo=bar"

overminder:
ping

alexcrichton:
Thanks!

PR #93, opened 2016-01-15
Increase timeout of request in test

msierks:
I noticed the **test_keep_alive::test_post_get_requests** test was periodically failing on travis. Increasing the timeout should should fix that.


alexcrichton:
I've got a rewrite over at https://github.com/alexcrichton/curl-rust/pull/105 which I think should obviate the need for this, but feel free to take a look!

msierks:
I will when I get a chance.

PR #95, opened 2016-01-17
Add user-agent setter

matthiasbeyer:
This is my first Rust PR, so please point out what I can do better.

This implements a user-agent setter on the `Handle` type.

### DIFF
--- a/src/http/handle.rs
+++ b/src/http/handle.rs
@@ -93,6 +93,11 @@ impl Handle {
         self
     }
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Hey @matthiasbeyer , I'm taking over the curl crate for now and I've got an API rewrite over at https://github.com/alexcrichton/curl-rust/pull/105 which I believe should address this as well. If you'd like to take a look though feel free!

PR #102, opened 2016-04-25
Provide custom header callback in addition to progress

xen0n:
This feature seems necessary for implementing rust-lang-nursery/rustup.rs#344, as `curl_easy_perform` is blocking.

I particularly don't like the name for the new builder method `header_cb` because it clashes with `headers`, but I couldn't immediately come up with a better name. Any suggestions on this?

BTW I'm relatively new to Rust (haven't used it for almost 1 year), so please kindly point out any mistake I might have made. Thanks!

### DIFF
--- a/src/ffi/easy.rs
+++ b/src/ffi/easy.rs
@@ -11,10 +11,16 @@ use http::{header, Response};
 
 use curl_ffi as ffi;
 
+pub type HeaderCb<'a> = FnMut(&str, &str) + 'a;
 pub type ProgressCb<'a> = FnMut(usize, usize, usize, usize) + 'a;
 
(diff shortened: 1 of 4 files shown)


alexcrichton:
Hey @xen0n , I'm taking over the curl crate for now and I've got an API rewrite over at https://github.com/alexcrichton/curl-rust/pull/105 which I believe should address this as well. If you'd like to take a look though feel free!

xen0n:
@alexcrichton I took a glimpse and it's an elegant API! Very close to C API and almost every feature is exposed. I will close this PR favoring the API you suggested instead.

Great work! :smiley:

PR #103, opened 2016-04-28
Upgrade to rust-url 1.0

SimonSapin:
Since `Url` is (indirectly) part of the public API, I think this is a breaking change.

It would be possible to depend on `url = ">= 0.5, < 2.0"` (so that this wouldn’t be a breaking change) if `as_str` is reverted to `to_string`, but that would leave an unnecessary `String` allocation.

`test_proxy` fails for me locally with "Timeout was reached", but it also does on master.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ repository = "https://github.com/carllerche/curl-rust"
 description = "Rust bindings to libcurl for making HTTP requests"
 
 [dependencies]
-url = "0.5.0"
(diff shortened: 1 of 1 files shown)


alexcrichton:
Hm I'd prefer to hold off on bumping the major version just yet, I've got a rewrite queued up already for this and ideally that'd land at the same time

alexcrichton:
Perhaps this could add the larger version range for now?

SimonSapin:
Updated the PR and filed #104.

PR #113, opened 2016-06-10
Fix compilation on Solaris/Illumos

nbaksalyar:
This fixes `curl-sys` compilation on Solaris, which requires explicit link with `libnls` for `inet_ntop` and `inet_pton` symbols.


alexcrichton:
Thanks!

PR #114, opened 2016-06-19
Hotfix/README

waldemarnt:
Hey buddy.
I faced some problems while implementing some requests, like this:
screen shot 2016-06-18 at 9 44 36 pm
That happens because there's no response for read_function in the README POST example.
I just fixed the example adding this information in this commit https://github.com/alexcrichton/curl-rust/commit/49c715d70644ca2dbe11126aa43fca926180eca6

The second stuff that I faced was about the form. My server had not received any data because I had not sent the correct size. This PR https://github.com/alexcrichton/curl-rust/pull/114/commits/cf5e1e98f0436ab352b566bfe5d58514c6dfb5b7 fix this information.

Thanks, your lib is very useful :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ fn main() {
     let mut easy = Easy::new();
     easy.url("https://www.rust-lang.org/").unwrap();
     easy.write_function(|data| {
-        stdout().write(data).unwrap()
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #134, opened 2016-10-30
Fix incorrect CURLOPT doc for `progress_function`

simias:
The method is implemented using CURLOPT_PROGRESS* callbacks, not CURLOPT_XFERINFO*


alexcrichton:
Ah yes indeed, thanks!

PR #135, opened 2016-10-30
Implement a safe interface to CURLOPT_SSL_CTX_FUNCTION and _DATA.

simias:
I'm currently trying to port the LastPass command line client from C to Rust and it uses this callback to implement certificate pining: https://github.com/lastpass/lastpass-cli/blob/master/http.c#L292

Given that this is a pretty security-critical piece of code I was conservative with this patch but I'm not entirely satisfied with this Rust implementation from a "quality of life" standpoint.

The problem is that the documentation says that the `ssl_ctx` pointer is simply a pointer to the library's SSL_CTX. I'm not really sure what to do with that, can we assume that it's always an OpenSSL context? If so we could provide a safe `openssl::ssl::SslContext` instead.

The docs mention that it can be used with the "wolfSSL/CyaSSL" libraries as well, but then again I've seen other [hardcoded calls to openssl](https://github.com/alexcrichton/curl-rust/blob/master/src/lib.rs#L93) elsewhere in curl-rust so I'm not really sure where we stand here.

The lastpass client seems to assume it's an OpenSSL context at least...

### DIFF
--- a/src/easy.rs
+++ b/src/easy.rs
@@ -119,6 +119,7 @@ struct EasyData {
     debug: Option<Box<FnMut(InfoType, &[u8]) + Send>>,
     header: Option<Box<FnMut(&[u8]) -> bool + Send>>,
     progress: Option<Box<FnMut(f64, f64, f64, f64) -> bool + Send>>,
+    ssl_ctx: Option<Box<FnMut(*mut c_void) -> Result<(), Error> + Send>>,
     header_list: Option<List>,
     form: Option<Form>,
     error_buf: RefCell<Vec<u8>>,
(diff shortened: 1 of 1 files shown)


simias:
Ah, I forgot to mention that the ssl_ctx callback normally also has access to a `CURL *` pointer (like the debug callback). However I'm not sure if we can expose this safely to the rust code so I just hide this pointer like for the `debug_cb` implementation. I don't need it in my use case so it's good for me but it could be limiting in other situations.

src/easy.rs
```diff
@@ -2894,6 +2945,50 @@ extern fn easy_progress_cb(data: *mut c_void,
...
+    let result =
+        match result {
+            Some(r) => r,
+            None => Err(Error::new(curl_sys::CURLE_SSL_CONNECT_ERROR)),
+        };
+
+    match result {
+        Ok(()) => curl_sys::CURLE_OK,
+        Err(e) => e.code(),
+    }
```
alexcrichton: Perhaps this could just be:

```rust
match result {
    Some(Ok(())) => curl_sys::CURLE_OK,
    Some(Err(e)) => e.code(),
    None => curl_sys::CURLE_SYS_CONNECT_ERROR,
}
```

simias:
Thank you for the feedback.

I hope I haven't forgotten anything this time.

alexcrichton:
Thanks!

PR #136, opened 2016-12-24
Use metadeps to specify pkg-config dependencies declaratively

joshtriplett:
This makes it easier for distribution packaging tools to generate
appropriate package dependencies.


joshtriplett:
The single appveyor error here seems completely unrelated and pre-existing; it occurs on current master as well. Everything else seems to work fine.

alexcrichton:
Closing for now as I think the metadeps story is still in the works

PR #137, opened 2016-12-26
curl-sys: fix call to `configure` in build script

japaric:
the old form caused `configure` to probe the environment using some
tests that looked like this:

```
$ $TARGET-gcc conftest.c
$ ./conftest
```

which always failed because `conftest` could have been compiled for e.g.
MIPS but was being called from a x86_64 system.

With the new form, `configure` never call `conftest` in its tests.

r? @alexcrichton

### DIFF
--- a/curl-sys/build.rs
+++ b/curl-sys/build.rs
@@ -143,9 +143,20 @@ fn main() {
 
     if target != host &&
        (!target.contains("windows") || !host.contains("windows")) {
-        cmd.arg(format!("--build={}", host));
-        cmd.arg(format!("--host={}", target));
-        cmd.arg(format!("--target={}", target));
+        // NOTE GNU terminology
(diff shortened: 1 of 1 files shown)


japaric:
See [this comment/thread](https://github.com/japaric/cross/issues/4#issuecomment-269242262) for more background information.

alexcrichton:
Looks good to me, thanks! I think the AppVeyor failure is benign but the Travis one looks legit?

japaric:
restorted the old host/target flags for windows targets.

r? @alexcrichton

alexcrichton:
Thanks!

japaric:
Thanks, @alexcrichton. It'd be great if you could release a new curl-sys version with this change.

alexcrichton:
Certainly, done!

japaric:
Thank you, @alexcrichton!

PR #140, opened 2017-01-08
Enable using precompiled mingw version.

crlf0710:
Hi, I tried building this repo yesterday and found that it failed to build. 

Yes i know that it's because i'm building under Windows and yes i've building it from cmd.exe, not mingw64 shell. 

However after investigation i found that there's a line precluding using the mingw libcurl prebuilt version.
I edited it and it successfully use the information retrieved from pkg-config, which builds fine.

Is there a special reason for not allowing using the mingw version?

### DIFF
--- a/curl-sys/build.rs
+++ b/curl-sys/build.rs
@@ -34,7 +34,7 @@ fn main() {
     }
 
     // Next, fall back and try to use pkg-config if its available.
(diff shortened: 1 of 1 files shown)


crlf0710:
Sorry, just found that there's an existing PR to do so. I'll just close this.

PR #145, opened 2017-01-25
Changes to support HTTP/2

maxdymond:
Allow HTTP version to be set on an Easy object.
Allow linking to nghttp2 when compiling from scratch.

### DIFF
--- a/curl-sys/build.rs
+++ b/curl-sys/build.rs
@@ -68,6 +68,7 @@ fn main() {
 
     let openssl_root = register_dep("OPENSSL");
     let zlib_root = register_dep("Z");
+    let nghttp2_root = register_dep("NGHTTP2");
 
     let cfg = gcc::Config::new();
     let compiler = cfg.get_compiler();
@@ -146,10 +147,15 @@ fn main() {
         cmd.arg(format!("--target={}", target));
     }
 
+    if let Some(root) = nghttp2_root {
+        cmd.arg(format!("--with-nghttp2={}", msys_compatible(&root)));
+    } else {
+        cmd.arg("--without-nghttp2");
+    }
+
     cmd.arg("--without-librtmp");
     cmd.arg("--without-libidn");
     cmd.arg("--without-libssh2");
-    cmd.arg("--without-nghttp2");
     cmd.arg("--disable-ldap");
     cmd.arg("--disable-ldaps");
     cmd.arg("--disable-ftp");
--- a/curl-sys/lib.rs
+++ b/curl-sys/lib.rs
@@ -605,6 +605,22 @@ pub const CURL_IPRESOLVE_WHATEVER: c_int = 0;
 pub const CURL_IPRESOLVE_V4: c_int = 1;
 pub const CURL_IPRESOLVE_V6: c_int = 2;
 
+/// These enums are for use with the CURLOPT_HTTP_VERSION option.
+pub type curl_HttpVersion = __enum_ty;
+/// Setting this means we don't care, and that we'd like the library to choose 
+/// the best possible for us!
+pub const CURL_HTTP_VERSION_NONE: curl_HttpVersion = 0;
+/// Please use HTTP 1.0 in the request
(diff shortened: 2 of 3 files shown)


maxdymond:
Hmm. I have just noticed that this project only claims support with Curl 7.24.0, whereas I've been testing with Curl 7.52.1. Feel free to reject this pull-request based on that, if necessary.

alexcrichton:
Nah this looks great to me, thanks! I wonder if perhaps this should be behind a feature flag though disabled by default as it's relatively recent? The CI builds look like they may all be failing as well :(

maxdymond:
Yeah - I'm a bit confused by that as I basically copied existing similar code. It looks like it can't find curl_HttpVersion, but that type is defined as an __enum_ty.

I'll try and gate this behind a feature flag. I'm pretty new to Rust!

alexcrichton:
I wonder if it may be an older version of libcurl? I haven't checked the versions and/or headers to see if this matches up.

If you need any help though just let me know

alexcrichton:
Hm so actually, on second though, we can probably avoid the Cargo feature here. In general curl is very good about back and future compatibility b/c everything goes through setopt which should return errors about unknown options. In that sense we could be sure to run `systest` (verifying FFI bindings) against recent versions of curl and don't worry about older versions as they should all be backcompat.

That way if you enable http/2 but it's not supported then curl'd just return a normal error perhaps. How's that sound?

maxdymond:
Sounds ok to me. I guess if curl setopt is passed a value that it doesn't know about and returns an error in that case, then we can define whatever we like on our side. CURL_HTTP_VERSION is present as an option in 7.24, so no problems there.

Is there anything else you'd want me to change, other than removing the gating that I added? I'm mindful of the fact that I've not yet added any tests for this feature, mostly because I'm not sure how to invoke them outside of the CI process.

alexcrichton:
Nah looks good to me! I'm ok eliding tests here as there's not a huge amount to test anyway. The "systest" crate already validates the correctness of all the type/constant definitions, so I'm not too worried there.

maxdymond:
Ok, so removing the gating causes systest to fail. Probably because the constants CURL_HTTP_VERSION_2TLS etc are not defined in curl.h, which seems reasonable.

So, is there a way of getting systest to not worry about these values not being present?

alexcrichton:
You can add constants to ignore in `systest/build.rs` through [`skip_const`](https://docs.rs/ctest/0.1.1/ctest/struct.TestGenerator.html#method.skip_const) perhaps?

maxdymond:
Ok, I think this is a happy medium. I've gated the HTTP_VERSION_2 constants behind the "http2" feature in the systest package - this means that by default, it should just compile as normal and test the V1 constants. If you want to test the V2 constants, you can do `cargo run --features=http2`. Let's see if Travis is happy with it..

alexcrichton:
Looks great, thanks! Could you modify one of the images where curl is built from source to pass the `--features http2` flag to verify the constant?

maxdymond:
Will that work? It'll only work if the version of curl in the image is >= 7.49.0. By default I'm not sure what versions the images are running with!

alexcrichton:
Yeah in theory it should. The bundled version in curl-sys should be very up-to-date

maxdymond:
Ok, if you point me at where I should be editing to affect the images, I'll try and make a change tomorrow morning.

alexcrichton:
The actual script to run tests is at [`ci/run.sh`](https://github.com/alexcrichton/curl-rust/blob/master/ci/run.sh) and that [folder](https://github.com/alexcrichton/curl-rust/tree/master/ci) contains a number of docker images that tests are run inside. Perhaps you could add an env var indicating "run these tests" and add that to the linux64 docker image?

maxdymond:
Ok, that all seems to have worked, and everything except the broken AppVeyor test is working. I think we might be there? 

Do you want me to squash this into a single commit?

alexcrichton:
Looks great! No need to squash.

I need to get around to fixing that AppVeyor test, but definitely not your problem.

PR #150, opened 2017-02-04
Add getters for time

simonvandel:
Adds bindings for the following options.
- CURLINFO_TOTAL_TIME
- CURLINFO_NAMELOOKUP_TIME
- CURLINFO_CONNECT_TIME
- CURLINFO_APPCONNECT_TIME
- CURLINFO_PRETRANSFER_TIME
- CURLINFO_STARTTRANSFER_TIME
- CURLINFO_REDIRECT_TIME

Is there a better way than transmute in getopt_double?

### DIFF
--- a/src/easy.rs
+++ b/src/easy.rs
@@ -14,6 +14,7 @@ use std::path::Path;
 use std::slice;
 use std::str;
 use std::time::Duration;
+use std::mem::transmute;
 
 use curl_sys;
 use libc::{self, c_long, c_int, c_char, c_void, size_t, c_double, c_ulong};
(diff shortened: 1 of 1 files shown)


src/easy.rs
```diff
@@ -2864,6 +2951,11 @@ impl Easy {
         }
     }
 
+    fn getopt_double(&mut self, opt: curl_sys::CURLINFO) -> Result<c_double, Error> {
+        self.getopt_ptr(opt)
+            .map(|p| unsafe {transmute::<_, c_double>(p)})
```
alexcrichton: This looks to be failing on 32-bit platforms, perhaps this could just call the FFI fn directly?

simonvandel: I am not really sure what you mean by calling the FFI function directly.
How does one go from *const i8 to f64 in a portable way?

simonvandel:
I agree on Duration being more rustic, implemented in newest commit.

simonvandel:
I added a `as u64` cast on the pointer which at least makes it compile. I haven't tested whether the returned value is sensible on 32bit though!

src/easy.rs
```diff
@@ -2864,6 +2958,11 @@ impl Easy {
         }
     }
 
+    fn getopt_double(&mut self, opt: curl_sys::CURLINFO) -> Result<c_double, Error> {
+        self.getopt_ptr(opt)
+            .map(|p| unsafe {transmute::<_, c_double>(p as u64)})
```
alexcrichton: I think we'll still want to call the raw FFI function directly here b/c we want 64-bits of information back, but on 32-bit this is only giving us 32-bits of information back

simonvandel:
The failing test on AppVeyor should be unrelated. Do you want me to squash the two commits?

alexcrichton:
Nah looks good to me, thanks!

simonvandel:
I'm working on a project using tokio_curl and this library. Could you update the crates.io versions of these libraries so I can remove my local forks?

alexcrichton:
Certainly!

PR #151, opened 2017-02-07
Add Multi::wait function

lluchs:
[curl_multi_wait](https://curl.haxx.se/libcurl/c/curl_multi_wait.html) is the recommended function for performing requests with the multi API.

I'm not sure about the `i32` timeout parameter. While it matches the curl API, a `Duration` would probably be nicer. These have a larger range though, so we'd either have to return an error or clamp to the largest `i32` for huge timeout values.

`curl_multi_wait` was introduced five years ago in curl 7.28.0 which is probably the minimum required curl version then.

### DIFF
--- a/curl-sys/lib.rs
+++ b/curl-sys/lib.rs
@@ -9,7 +9,7 @@ extern crate openssl_sys;
 #[cfg(windows)]
 extern crate winapi;
 
-use libc::{c_int, c_char, c_uint, c_long, c_double, c_void, size_t, time_t};
+use libc::{c_int, c_char, c_uint, c_short, c_long, c_double, c_void, size_t, time_t};
 use libc::c_ulong;
 
 #[cfg(unix)]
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks! Could the extra fds argument also be bound? And yeah ideally we'd take a Duration here and just return errors and/or clamp for too-long duration.s

lluchs:
Okay, I changed it to take a Duration and to clamp at the maximum value. (I wish the conversion was easier!)

I don't know what the Rust side of the extra_fds argument would look like. Just passing the `curl_waitfd` struct through wouldn't make a nice API, I guess. Rust doesn't really provide a pollfd abstraction we could use, does it?

I suspect that Rust users won't have any unrelated raw file descriptors anyways. Even the tests use mio instead of working with the raw Unix APIs.

alexcrichton:
Ah yeah we wouldn't want to expose the raw FFI type, but I think it's useful to expose to ensure we don't need to add a second API that does work with this in the future. I'd imagine something like:

```rust
struct WaitFd {
    inner: curl_wait_fd,
}

impl WaitFd {
    // accessors/getters/constructors
}

fn wait(&self, waitfds: &mut [WaitFd], timeout: Duration) -> ...
```

lluchs:
Thanks for the hint, I implemented that now.

I'm not sure about the safety of the `waitfds.as_mut_ptr() as *mut curl_sys::curl_waitfd` slice cast. It works fine, but people on IRC were inconclusive about whether this is undefined/unspecified behavior.

There's also a potential overflow in the `waitfds.len() as u32` cast, but you can't have that many file descriptors anyways.

alexcrichton:
Looks great to me, thanks!

alexcrichton:
Oh it looks like travis is failing?

lluchs:
You're right, sorry. I don't have macOS, but I guess `poll(2)` on Darwin works slightly different than on Linux in how it sets `pollfd.revents`? I just pushed a commit that limits that test to Linux only, which should be enough to verify that the Rust wrapper functions are working correctly.

alexcrichton:
Sounds unusual! Limiting the test though seems fine to me, thanks again for the PR!

Issue #144, opened 2017-01-23
Support for curl options?


tomyan:
I'd like to make use of https://curl.haxx.se/libcurl/c/CURLOPT_UNIX_SOCKET_PATH.html for making requests over a domain socket. I have tried to use `req.setopt_str(...)`, but that doesn't work as it's private. Is there a way to do this already, or otherwise something you'd appreciate a PR for (pretty new to Rust, so might take a while)? I also couldn't see a way to import the `CURLOPT_UNIX_SOCKET_PATH` constant - maybe that needs to be added to the curl_sys crate?

Thanks

Tom

alexcrichton:
Ah yeah those helpers are just private functions, but no option is left out explicitly! Want to send a PR implementing this option?

alexcrichton:
Unix socket support added in https://github.com/alexcrichton/curl-rust/pull/154, so closing.

PR #154, opened 2017-03-09
Unix domain socket support

kstrempel:
The long missing unix domain socket support for curl-rust. :-) 

Related to #144 (Support for curl options?)

Currently I don't know how to write a test for this repro. I tested it on my machine with OSX against the docker socks api.


src/easy.rs
```diff
@@ -463,6 +463,19 @@ impl Easy {
...
     }
 
+    /// Provides the unix domain socket which this handle will work with.
+    ///
+    /// The string provided must be unix domain socket -encoded with the format:
+    ///
+    /// ```text
+    /// /path/file.sock
+    /// ```
+    pub fn unix_socket(&mut self, unix_domain_socket: &str) -> Result<(), Error> {
```
alexcrichton: Perhaps this could take the same argument as `CString::new`, which is `T: Into<Vec<u8>>`?

alexcrichton:
Thanks! It looks like there may be CI failures though?

alexcrichton:
Thanks!

PR #156, opened 2017-03-15
Explicitly disable PSL support for curl-sys

inejge:
If the PSL (Public Suffix List) library is installed on the host, curl's `configure` script will enable it, but the curl-sys `build.rs` doesn't include it in the library list, so subsequent linking with libcurl-sys will fail. Since curl-sys doesn't need PSL support, it's best to disable it explicitly.


alexcrichton:
Thanks!

PR #178, opened 2017-07-20
added support for CURLOPT_RESOLVE

qfire:
I use this option to connect to a specific IP address of a multi-homed device but still be able to verify the certificate presented.

### DIFF
--- a/src/easy/handler.rs
+++ b/src/easy/handler.rs
@@ -1713,6 +1713,30 @@ impl<H> Easy2<H> {
         self.setopt_long(curl_sys::CURLOPT_IPRESOLVE, resolve as c_long)
     }
 
+    /// Specify custom host name to IP address resolves.
+    ///
+    /// Allows specifying hostname to IP mappins to use before trying the
+    /// system resolver.
+    ///
+    /// # Examples
+    /// ...
+    /// use curl::easy::{Easy, List};
+    ///
+    /// let mut list = List::new();
+    /// list.append("www.rust-lang.org:443:192.241.171.49").unwrap();
+    ///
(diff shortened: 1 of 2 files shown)


src/easy/handler.rs
```diff
@@ -1713,6 +1713,30 @@ impl<H> Easy2<H> {
...
+    /// list.append("www.rust-lang.org:443:192.241.171.49").unwrap();
+    ///
+    /// let mut handle = Easy::new();
+    /// handle.url("https://www-rust-lang.org/".unwrap();
+    /// handle.resolve(list).unwrap();
+    /// handle.perform().unwrap();
+    /// ...
+    pub fn resolve(&mut self, list: List) -> Result<(), Error> {
+        let ptr = list::raw(&list);
+        self.inner.header_list = Some(list);
```
alexcrichton: I think we'll need a seprate storage slot for this `resolve` list because otherwise if you configure both headers and a resolve list they'll free the other's memory

src/easy/handler.rs
```diff
@@ -1713,6 +1713,30 @@ impl<H> Easy2<H> {
         self.setopt_long(curl_sys::CURLOPT_IPRESOLVE, resolve as c_long)
     }
 
+    /// Specify custom host name to IP address resolves.
+    ///
+    /// Allows specifying hostname to IP mappins to use before trying the
+    /// system resolver.
+    ///
+    /// # Examples
+    /// ...
```
alexcrichton: I think this'll need to be three backticks

alexcrichton:
Thanks!

PR #180, opened 2017-08-01
Allow OSX users to link with customly built libcurl

overminder:
This commit changes curl-sys's build script behavior for OSX targets
when environment variable RUST_CURL_DONT_USE_OSX_SHIPPED is set: it will
fall back to use pkg-config for library searching, which allows the
user to specify a custom libcurl.

This is useful in situations where the default system libcurl doesn't
provide enough functionalities (e.g. the user wants to use OpenSSL
rather than SecureTransport as the TLS backend).

### DIFF
--- a/curl-sys/build.rs
+++ b/curl-sys/build.rs
@@ -25,8 +25,8 @@ fn main() {
     let windows = target.contains("windows");
 
     // OSX ships libcurl by default, so we just use that version
-    // unconditionally.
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! Could you add a test for this on CI?

alexcrichton:
Closing due to inactivity for now

Issue #189, opened 2017-10-23
Build failure with MinGW


Bobo1239:
`cargo build` fails with
```
failed to open E:/Development/Rust/curl-rust/curl-sys/curl/lib/E:/Development/Rust/curl-rust/curl-sys/curl/lib/amigaos.c at E:/Development/Rust/curl-rust/curl-sys/curl/lib/checksrc.pl line 259.
```
using MinGW 64-bit (msys2).

Could we just remove (or restrict to msvc) [this check](https://github.com/alexcrichton/curl-rust/blob/f4a6e2fbf8885216f2d942f86e291cea68c78a56/curl-sys/build.rs#L39)? If I comment that out my build succeeds.

alexcrichton:
Can you gist the full log of the build failure?

Bobo1239:
Sure: https://gist.github.com/Bobo1239/bde296b5d1251b4634426bc2b68046c1

alexcrichton:
Ah the bug is [here](https://github.com/alexcrichton/curl-rust/blob/f4a6e2fbf8885216f2d942f86e291cea68c78a56/curl-sys/build.rs#L245-L252), it only handles builds on `C:\`, not on `E:\`, alas!

PR #190, opened 2017-10-27
Fix MinGW compilation on drives besides C:

Bobo1239:
Also removed some unnecessary (as far as I can see) indirection.

Fixes https://github.com/alexcrichton/curl-rust/issues/189.

### DIFF
--- a/curl-sys/build.rs
+++ b/curl-sys/build.rs
@@ -3,10 +3,11 @@ extern crate pkg_config;
 extern crate vcpkg;
 extern crate cc;
 
+use std::ascii::AsciiExt;
 use std::env;
 use std::ffi::OsString;
 use std::fs;
-use std::path::{PathBuf, Path};
+use std::path::{PathBuf, Path, Component, Prefix};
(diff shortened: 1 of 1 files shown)


src/multi.rs
```diff
@@ -341,10 +341,6 @@ impl Multi {
     /// error code from the transfer or just the fact that a transfer is
     /// completed. More details on these should be written down as well.
     pub fn messages<F>(&self, mut f: F) where F: FnMut(Message) {
-        self._messages(&mut f)
-    }
-
-    fn _messages(&self, mut f: &mut FnMut(Message)) {
```
alexcrichton: Oh could this be left as-is? The intention here is to reduce monomorphization burden on callers

alexcrichton:
Looks great, thanks!

src/multi.rs
Bobo1239: Ah I see, TIL. Removed the second commit.

alexcrichton:
Thanks!

PR #191, opened 2017-11-02
Replace --without-libidn with --without-libidn2

mneumann:
Curl uses libidn2, libidn is gone. The `--without-libidn` has no effect.

I found this while building statically-linked binaries of cargo for
DragonFly (which fails without this option).


alexcrichton:
Thanks!

PR #192, opened 2017-11-21
add support for LIBCURL_LIB_DIR and LIBCURL_INCLUDE_DIR

qfire:
This patch allows specifying where to find the system libcurl and includes to link against.

In my use case this allowed me to use the x86_64-unknown-linux-musl target to statically link against a version of libcurl with features I needed.  The extra static dependency libraries I was able to specify in my project's own build.rs, so I didn't feel a need to support another variable to specify those.

### DIFF
--- a/curl-sys/build.rs
+++ b/curl-sys/build.rs
@@ -25,6 +25,31 @@ fn main() {
     let dst = PathBuf::from(env::var_os("OUT_DIR").unwrap());
     let windows = target.contains("windows");
 
+    println!("cargo:rerun-if-env-changed=LIBCURL_LIB_DIR");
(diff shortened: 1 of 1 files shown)


alexcrichton:
I think this is a use case where [overriding build scripts](http://doc.crates.io/build-script.html#overriding-build-scripts) is what you want?

qfire:
Unless I'm missing something that'll work for my local project to specify the libcurl dependency libraries when static linking but this patch to curl-sys is still needed to get curl-sys to link against the local custom libcurl itself.

The patch is similar to how other crates allow users to do this, for example openssl-sys or gpgme.

alexcrichton:
Perhaps yes, but that's also sort of because overrides aren't that well known. For this crate, though, that's the way to go for overriding it.

PR #198, opened 2018-03-02
 added function to check if time condition was met 

KaKnife:


alexcrichton:
Thanks! Since `bool` doesn't have a great ABI definition, could this use `getopt_long` in the implementation?

KaKnife:
Is that what you had in mind?

alexcrichton:
Indeed, thanks!

PR #203, opened 2018-03-09
Allow OSX users to link with customly built libcurl (again)

hhatto:
continuation of #180

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
       env: TARGET=x86_64-apple-darwin NO_ADD=1
     - os: osx
       rust: stable
-      env: TARGET=x86_64-apple-darwin NO_ADD=1 HOMEBREW=1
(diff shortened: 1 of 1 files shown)


hhatto:
I'm tested on [this branch](https://github.com/hhatto/curl-rust/tree/test-ci). ([changes](https://github.com/hhatto/curl-rust/compare/continuation-pr-180...hhatto:test-ci))

But, version of Homebrew's curl is 7.58.0, now.
This is `CURLVERSION_NOW==CURLVERSION_FIFTH` and `sturct curl_version_info_data` also has a difference.
As a result, systest fails. ([travis-ci's build result](https://travis-ci.org/hhatto/curl-rust/builds/351715059))

I would like opinions as to whether there is any good way to do it.

alexcrichton:
Ah yeah we test with a whole slew of versions on CI so unfortunately I don't think we'll be able to add a test per se about what version is detected. Although perhaps the CI matrix could set the expected version, the test could read curl's version and the env var, and then assert the two are the same?

alexcrichton:
Closing due to inactivity for now, but feel free to resubmit with a travis test!

PR #204, opened 2018-03-26
Bump winapi to 0.3

Eijebong:


Eijebong:
Just a heads up, I'm going blind there and trusting CI as I don't have a windows machine and couldn't get cross compilation to work

Eijebong:
My code seems to be right. Why is it still trying to compile winapi 0.2 ?

alexcrichton:
Thanks for the PR! For now though this is a backwards-incompatible change (unfortunately) so I'm gonna hold off on this for now as I'm not quite ready to make a breaking change to the `curl` crate.

Eh2406:
In what way is this a breaking change?

Eh2406:
I forgot that most of the `winapi` api did not move, and did not see `pub` in the diff. Sorry for the noize.

infinity0:
When can we expect this to land? The function linked to, `fd_set` is indeed exposed but it seems all that changed between winapi 0.2 and 0.3 was that it was moved to a more inner crate, but any crate that re-exports it with a `pub use` should effectively be unchanged?

alexcrichton:
The next major version of this crate can drop the `winapi` dependency, but until then it's unfortunately stuck :(

Is there a pressing reason though to upgrade?

infinity0:
We're trying to avoid packaging winapi 0.2 in Debian and this is the only dependency of cargo that still uses it. We don't actually use winapi but keep the dependencies around to avoid patching the many crates that depend on them. I suppose we can work around it by applying this patch to the existing version for now, since nothing currently in Debian is trying to cross-compile rust packages to windows.

PR #205, opened 2018-04-02
fix(cargo): Update minimal dependency versions so that 'cargo build -Z minimal-versions' works

klausi:
The create does not compile with `cargo build -Z minimal-versions` because a lot of minimal versions are underspecified. I bumped them until I could build successfully.

Ideally we should also add something like `cargo update -Z minimal-versions && cargo build` to the Travis CI config so that this does not happen again in the future. I'm not sure where to put that exactly?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ appveyor = { repository = "alexcrichton/curl-rust" }
 [dependencies]
 libc = "0.2"
 curl-sys = { path = "curl-sys", version = "0.4.1" }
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks so much!

Sounds like a great idea to me to add this to CI eventually (and I think long-term we'd even do this on `cargo publish`). For the time being though I think it's ok to skip that as it's still a nightly Cargo feature and otherwise this lib isn't changing a whole lot

PR #206, opened 2018-04-10
fix(cargo): Install highest possible libz-sys dependency version

klausi:
Follow-up to #205 .

Sorry, I thought that "0.1.2" would allow any later version, but cargo interprets that as 0.1.x only. We want libz-sys 1.x also.


alexcrichton:
Aha oops yes indeed!

PR #208, opened 2018-04-14
Update rust-curl to winapi 0.3

matklad:
Note that curl-sys already uses 0.3


matklad:
Looks like a dupe of https://github.com/alexcrichton/curl-rust/pull/204 :)

alexcrichton:
Ah yes this is indeed a dupe and it's also a [public dependency](https://github.com/alexcrichton/curl-rust/pull/204#issuecomment-378023790) so I'd like to avoid having a breaking change for now.

PR #209, opened 2018-04-20
Manifest: add excludes

ishitatsuyuki:
These are two of the biggest data directories that get extracted to Cargo cache, but they are not actually useful. Exclude them when publishing.


alexcrichton:
Actually this breaks `cargo package` (it's also on the wrong manifest), so I reverted.

PR #212, opened 2018-06-24
Forwarding impl for Handler

ishitatsuyuki:


ishitatsuyuki:
Closing: this was originally intended for tokio-curl, but I've came up with a different approach that doesn't need forwarding.

PR #213, opened 2018-06-25
unsafe impl Send for Multi

ishitatsuyuki:


alexcrichton:
Hm I vaguely remember intentionally not doing this, although I'm forgetting now if it was being conservative or because it was unsound. Is it perhaps possible to implement https://github.com/alexcrichton/tokio-curl/pull/21 without this?

ishitatsuyuki:
I don't think so; the current Windows implementation uses `Multi` on another thread (which is broken by the previous PR already), and it's not possible to do that without sending that.

Though, I'll try to figure out in what way this may be unsound.

ishitatsuyuki:
So what I've found is only that `Easy2` or `Handler` doesn't enforces the `Send` bound, as well as the `add2` method. Apart from that, I think there shouldn't be any pitfalls in the curl library itself.

alexcrichton:
Hm ok I think that this may actually be unsound due to how types like `EasyHandle` are mananged. If a `Multi` is on one thread and an `EasyHandle` is dropped on another thread, I believe that causes issues?

ishitatsuyuki:
I guess we probably should make https://github.com/alexcrichton/tokio-curl/blob/13c13340232dbe8625d6787d78e24b5356b9a635/src/lib.rs#L137 take a `FnOnce() -> Multi` instead? This is a breaking change, though luckily we haven't released master so we don't need to mess with semver.

If you think that's a good idea, please close this.

alexcrichton:
I've had to revert https://github.com/alexcrichton/tokio-curl/commit/47b37f501d0b4d9512f4fddd4cbc9cee49be5570 as it didn't actually compile on Windows, so yes it'll need to be `FnOnce` or we'll need to figure out a different strategy altogether. As-is though I think this is unfortunately not sound.

PR #222, opened 2018-08-12
Add unpause methods for handles attached to multi

sagebind:
Expose the pause/unpause interface to easy handles even after they are attached to a multi handle. This allows you to suspend/resume individual transfers when using the multi interface (for example, if your callbacks are interacting with non-blocking I/O and receive `WouldBlock`).

Yeah, another thing I need for [cHTTP](https://github.com/sagebind/chttp)...

### DIFF
--- a/src/multi.rs
+++ b/src/multi.rs
@@ -708,6 +708,34 @@ impl EasyHandle {
                                              token))
         }
     }
+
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #229, opened 2018-09-24
Point docs for Transfer::perform to Easy::perform

bupknar:
The docs for `Transfer::perform` erroneously said that it was the same as `Easy::transfer`. This should point to `Easy::perform` instead.


alexcrichton:
Thanks!

PR #230, opened 2018-10-01
Make the openssl dependency optional (with default on).

adrian-budau:
For work-reasons we want to use curl without `openssl` support (mostly to save binary sizes). This pull requests just makes `openssl` optional under a feature with the same name.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,8 +39,8 @@ mio-extras = "2.0.3"
 members = ["systest"]
 
 [features]
-default = ['openssl']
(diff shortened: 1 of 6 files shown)


alexcrichton:
Thanks! This looks good to me, but perhaps it can be called `ssl` and it would disable ssl across all platforms if it's not enabled?

adrian-budau:
I implemented the changes. Currently the doc tests that hit `https://...` fail when running with --no-default-features. Is there anyway I could deactivate them when a certain feature is not on?

alexcrichton:
Thanks!

PR #233, opened 2018-10-27
Add support for enabling SPNEGO when building curl

codido:
On Linux, link against the MIT gssapi library. On Windows, use SSPI.


alexcrichton:
Thanks!

PR #244, opened 2019-01-31
Add support for configuring DNS servers

thijsc:
Disabled feature by default because this only works when curl is linked with `c-ares`.


sagebind:
Indeed, it looks like curl just returns `CURLE_NOT_BUILT_IN` if the configured DNS backend doesn't support it:

https://github.com/curl/curl/blob/master/lib/hostsyn.c#L59-L69

thijsc:
Correct, it gives an error:

```
A requested feature, protocol or option was not found built-in in this libcurl due to a build-time decision.
```

I've removed the feature and tweaked the doc comment a bit to make the requirement to use this more clear.

alexcrichton:
👍

thijsc:
Would you mind bumping the crate version so we can use this?

alexcrichton:
Sure thing, should be published now

PR #253, opened 2019-02-15
Support more ssl versions

### LINKED ISSUE #252: support more ssl versions
I see that tlsv1.1 and tlsv1.2 are commented out in the code, see the following line:

https://github.com/alexcrichton/curl-rust/blob/cf4982390b9a10dffd94f82f7b40ab073ad2c2f9/src/easy/handler.rs#L484

any specific reason? I'd like to extend this SslVersion to support more versions. If you're OK, I can help send a simple PR.

guanqun:
fix #252


guanqun:
hmm, when I test it locally on local master branch, the 'resolve' test case also fails...

alexcrichton:
No need to worry about that, thanks!

PR #260, opened 2019-03-29
Do not mention *_sys crates if they are not needed.

hvenev:
They are only used so that the corresponding native libraries are linked
in to the static cURL library. This is not needed when building against
the system library:
  - If it's a shared library, it is linked against them already and they
 are not a part of the public interface.
  - Both pkgconfig and vcpkg deal with these dependencies and provide
  the right link args to cargo.


alexcrichton:
Thanks!

PR #261, opened 2019-03-29
Remove unnecessary call to openssl_sys::init()

hvenev:
curl_global_init() alreary initializes OpenSSL correctly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,9 @@ matrix:
     - os: linux
       rust: stable
       env: TARGET=x86_64-unknown-linux-gnu DOCKER=linux64-curl NO_ADD=1
+    - os: linux
+      rust: stable
+      env: TARGET=x86_64-unknown-linux-gnu DOCKER=centos7 NO_ADD=1
     - os: osx
       rust: stable
       env: TARGET=x86_64-apple-darwin NO_ADD=1
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,26 @@
+use std::env;
+use std::str::FromStr;
+
+fn main() {
+    // OpenSSL >= 1.1.0 can be initialized concurrently and is initialized correctly by libcurl.
+    // <= 1.0.2 need locking callbacks, which are provided by openssl_sys::init().
+    match env::var("DEP_OPENSSL_VERSION") {
+        Ok(ver) => {
+            let ver = u32::from_str(&ver).unwrap();
+            if ver < 110 {
+                println!("cargo:rustc-cfg=need_openssl_init");
+            }
+        }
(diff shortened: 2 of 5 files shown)


alexcrichton:
Judging from https://github.com/alexcrichton/curl-rust/commit/86eeb2906e525176682e30981bfbe361bb1af04a, it looks like libcurl may not have always done this? It's probably best to leave it in unless it's harmful perhaps?

hvenev:
It has done it pretty much since the beginning - d300cf4d84182b111f6b364a89a9af00eca7f425 is the earliest record I could find.

One exception is when the system libcurl is built without OpenSSL (e.g. using GnuTLS, NSS, mbedTLS). In this case initializing OpenSSL makes even less sense.

This may break crates that use both curl and OpenSSL and rely on curl initializing OpenSSL in the case when curl is not built with openssl.

However, this PR, together with #260, can be used to effectively drop the dependency on openssl-sys.

alexcrichton:
I sort of distinctly remember though adding this in to fix issues that arose in the ecosystem. I don't currently have the time to dig those up again but unless there's a concrete reasons to do this I'd prefer to leave it as is

hvenev:
The problem is that `openssl_sys::init()` used to do quite a lot more: [openssl-v0.8.0:openssl-sys/src/lib.rs:505](https://github.com/sfackler/rust-openssl/blob/652326003cefe215dbfc838051e6114515cc5190/openssl-sys/src/lib.rs#L505)

alexcrichton:
Ok thinking more on this I believe this is also required for correctness. We need to make sure that for at least historical versions OpenSSL isn't concurrently initialized, but if one thread is using `openssl_sys::init` and another is in `curl_sys::init`, then they could both transitively initialize OpenSSL concurrently causing issues.

I think OpenSSL 1.1.0+ has fixes for all of this because it handles initialization itself and is threadsafe, but that's not necessarily entirely phased out yet.

For now I think we need to keep this, but it sounds like it's causing issues for you so maybe there's some other configuration we need to ensure that it's not pulled in when not necessary? Maybe we can detect the OpenSSL version and avoid calling it if it's above 1.1.0?

hvenev:
I added a build script that should detect the necessary conditions. I haven't tested the OpenSSL one because I don't have access to a machine with OpenSSL < 1.1.0 or LibreSSL. Can you test that?

We could also not use openssl-probe when the bundled libcurl and the system OpenSSL are used by building curl with CURL_CA_FALLBACK. That makes it tell OpenSSL to enable the default certs.

alexcrichton:
Looks good to me, thanks! For CURL_CA_FALLBACK, want to open a follow-up?

PR #265, opened 2019-04-12
README.md: add troubleshooting section with NSS SSL hint

TorstenScheck:
As discussed in the cargo-vendor context.
If you think that the hint is also too prominent here, I can create a separate troubleshooting file.

Thank you very much for creating and maintaining such crucial crates besides all your other work.


alexcrichton:
Thanks!

PR #266, opened 2019-04-17
Add HTTP2 errors, and curl-sys error consts up to curl 7_64_1.

fluffysquirrels:
Required for https://github.com/rust-lang/cargo/issues/6788.

References:
* [libcurl-errors man page](https://curl.haxx.se/libcurl/c/libcurl-errors.html)
* [curl.h error enum](https://github.com/curl/curl/blob/curl-7_64_1/include/curl/curl.h#L477-L606)

The names of the methods on `curl::Error` could be better but will probably do. Happy to hear suggestions.


alexcrichton:
Thanks for this!

To get past CI I think that [this is going to need to be updated](https://github.com/alexcrichton/curl-rust/blob/9b9d1192d081fad21a2e42994e14c8a06f7c38a3/systest/build.rs#L62-L106) since these constants I don't think are present in older versions of curl. The "mac static" build is going to fail (due to unrelated reasons), but the other system systest tests are failing I thinkl

fluffysquirrels:
OK, I'll take a look :)

systest/build.rs
```diff
@@ -60,6 +60,18 @@ fn main() {
     });
 
     cfg.skip_const(move |s| {
+        if version < 64 {
```
fluffysquirrels: I doubt the version here is exactly when these consts were introduced. I'm hoping that doesn't matter.

fluffysquirrels:
All but macOS_static are passing now so I think I've ignored the right consts.

alexcrichton:
Thanks!

PR #276, opened 2019-07-15
derive Clone for ProxyType

0kaguya:
Now there have no comfortable way to set proxy type for more than one handler since `Easy::proxy_type` takes the ownership of a `ProxyType` enum. Deriving a `Clone` trait for `ProxyType` ensures such code possible:
``` rust
let proxy = read_proxy_type();
let handler1 = Easy::new();
let handler2 = Easy::new();
handler1.proxy_type(proxy.clone());
handler2.proxy_type(proxy.clone());
```
It will be better to modify the `Easy::proxy_type` function itself. But deriving a `Clone` trait is not harmful.

### DIFF
--- a/src/easy/handler.rs
+++ b/src/easy/handler.rs
@@ -392,7 +392,7 @@ unsafe impl<H: Send> Send for Inner<H> {}
 
 /// Possible proxy types that libcurl currently understands.
 #[allow(missing_docs)]
-#[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


sagebind:
Maybe it is safe to derive `Copy` as well?

alexcrichton:
Indeed yeah, want to add `Copy` too? Should otherwise be good to go!

PR #283, opened 2019-08-21
Link against clang runtime on static builds on macOS

### LINKED ISSUE #279: macOS link error with static curl — missing ___isOSVersionAtLeast
Linking with the static feature on macOS 10.13 fails with the following error:

```
Undefined symbols for architecture x86_64:
  "___isOSVersionAtLeast", referenced from:
      _sectransp_connect_step1 in libcurl_sys-72a3db42ebafc5cb.rlib(sectransp.o)
      _sectransp_connect_step2 in libcurl_sys-72a3db42ebafc5cb.rlib(sectransp.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

The problem is that `rustc` links with the `-nodefaultlibs` flag. This prevents `cc` from linking with `/Library/Developer/CommandLineTools/usr/lib/clang/10.0.0/lib/darwin/libclang_rt.osx.a` (or whatever version of xcode you have), which is where `___isOSVersionAtLeast` is located.

This is caused by the use of things like `__builtin_available(macOS 10.13.4, iOS 11, tvOS 11, *)`.  Usually clang will statically strip out this builtin, which is why it normally works on 10.14. However, if you are building on an older version, it adds the call to `___isOSVersionAtLeast` at runtime.  Strangely, using a minor semver version (like `10.13.4` instead of `10.13`) also prevents it from being determined statically when building on 10.13 (presumably it defaults to `--mmacosx-version-min=10.13` which sets the floor to `10.13.0` regardless which minor version you are on).

Here's a reduced example project that will fail on 10.14 (by testing for 10.15).
[foo.zip](https://github.com/alexcrichton/curl-rust/files/3412906/foo.zip)

Hopefully that's all clear. Unfortunately I don't know how to fix this.  Presumably `rustc` uses `-nodefaultlibs` for a good reason.  It seems unlikely to me, but maybe there is a way to discover the correct location of `libclang_rt.osx.a` and include it in the link list?  A more draconian solution would be to unset `HAVE_BUILTIN_AVAILABLE`, but I think that would disable some important features (like TLS 1.3).  Perhaps Alex has some clever ideas?

badboy:
On OSX we need to link against the clang runtime, which is hidden in some non-default path.
We can get the path from `clang --print-search-dirs`.
Kudos to @ehuss for finding that workaround.

Fixes #279.

### DIFF
--- a/curl-sys/build.rs
+++ b/curl-sys/build.rs
@@ -493,9 +493,14 @@ fn curl_config_reports_http2() -> bool {
 }
 
 fn macos_link_search_path() -> Option<String> {
-    let output = Command::new("clang").arg("--print-search-dirs").output().ok()?;
+    let output = Command::new("clang")
+        .arg("--print-search-dirs")
+        .output()
+        .ok()?;
     if !output.status.success() {
-        println!("failed to run 'clang --print-search-dirs', continuing without a link search path");
+        println!(
+            "failed to run 'clang --print-search-dirs', continuing without a link search path"
+        );
         return None;
     }
 
--- a/curl-sys/build.rs
+++ b/curl-sys/build.rs
@@ -47,6 +47,17 @@ fn main() {
             .status();
     }
 
+    if target.contains("apple") {
+        // On (older) OSX we need to link against the clang runtime,
+        // which is hidden in some non-default path.
+        //
+        // More details at https://github.com/alexcrichton/curl-rust/issues/279.
+        if let Some(path) = macos_link_search_path() {
+            println!("cargo:rustc-link-lib=clang_rt.osx");
+            println!("cargo:rustc-link-search={}", path);
+        }
+    }
+
     let dst = PathBuf::from(env::var_os("OUT_DIR").unwrap());
     let include = dst.join("include");
     let build = dst.join("build");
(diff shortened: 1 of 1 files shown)


alexcrichton:
Nice! This is a bit simpler than I thought it would be.

Two requests for an update:

* Could this be updated to basically ignore errors? I'm not sure how all OSX systems would respond to this, and I'm not sure if it's necessarily required on all systems (on older ones I think the symbol wasn't linked). In that sense it seems like it may be a bit more conservative to ignore errors and try to catch more linker errors later as they come up.

* Could the comment be expanded a bit to include some information that #279  contains or link to that issue directly?

dmitmel:
@alexcrichton Could you please share the logs of failed checks or trigger a rebuild on github actions by amending the commit (with no changes) and force pushing that? I'd like to take a look, but unfortunately, it looks like I don't have access to them.

badboy:
Rebased and force-pushed /cc @dmitmel

dmitmel:
Well, it looks like there are no errors now. I can confirm that this PR indeed fixes #279 on my machine with macOS 10.13.

alexcrichton:
Seems reasonable!

PR #284, opened 2019-08-22
Static OpenSSL option for MacOS and Windows

rustysec:
#278 take 2!

provides feature to opt in to static openssl linkage.


rustysec:
i experimented locally with moving the current crate into a subfolder (`curl-base`) and then being able to use platform specific dependencies to turn the features on and off. there's just a one line `lib.rs` that re-exports the current crate. seems like i can get the same default behavior that way. it's a lot of re-organizing, not sure how you feel about that.

anyway, this whole effort may not be worth it once `mesalink` will build properly on windows.

alexcrichton:
Hm yeah that may be a bit overkill for this, and if `mesalink` works that'd be great if that could be used!

PR #285, opened 2019-08-23
enhance 'ssl_version' to be able to set max version as well

guanqun:
Also fix a small typo in the comment.


sagebind:
Perhaps there is a backwards-compatible way to add this functionality instead? Maybe a separate `ssl_version_range` method?

guanqun:
Good suggestion, but ideally the interface exposed should be minimal.

guanqun:
do you think `ssl_min_max_version` is better than `ssl_version_range()`? @sagebind

guanqun:
do you mind bumping up a version for this small feature?

guanqun:
oh. i have another PR to add the copy and clone for sslversion. after that, I'm okay for bumping the version.

PR #290, opened 2019-08-27
add Clone/Copy to SslVersion

guanqun:


guanqun:
actually there are other enums missing these two derives. do you thin we should add them all?

guanqun:
Sure. Created a PR here: https://github.com/alexcrichton/curl-rust/pull/291

PR #291, opened 2019-08-27
add Clone/Copy for more enums

guanqun:


guanqun:
Can I ask for a version bump when this is merged?

I'd like to use the new `ssl_min_max_version()` API that gets recently merged.

PR #294, opened 2019-08-28
Remove `winapi 0.2` dependency

alexcrichton:
This commit removes the `winapi 0.2` dependency and also deletes the
`Multi::fdset` method. While technically a breaking change this was
deprecated 2 years ago at this point so I'm hoping that this crate can
practically get away with the breakage.


nox:
Why not just make a major bump if it's a breaking change… ?

PR #296, opened 2019-09-07
Getters on easy handles when attached to multi

### LINKED ISSUE #273: Need to Be able to borrow an Easy2 from a Multi
When a transfer is in progress, once you get the headers, but before the payload is received, there's no way to call functions such as `Easy2::response_code()`. This is because `Multi` takes the `Easy2` object and then there's no way to access it.

I suppose simply adding something like `fn Easy2Handle::get(&self) -> { &self.easy }` would be sufficient and safe.

sagebind:
Expose read-only getters that are safe to invoke while an easy handle is attached to a multi handle.

Also allow users to access the underlying curl pointer if the safe interface is not flexible enough for a given use-case.

Fixes #273.

### DIFF
--- a/src/multi.rs
+++ b/src/multi.rs
@@ -10,7 +10,7 @@ use libc::{c_char, c_int, c_long, c_short, c_void};
 #[cfg(unix)]
 use libc::{pollfd, POLLIN, POLLOUT, POLLPRI};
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
👍

PR #302, opened 2019-09-27
Re-export `spnego` feature from curl-sys

kathampy:


alexcrichton:
Thanks! I think though there's a build failure on CI?

kathampy:
> Thanks! I think though there's a build failure on CI?

I'm not sure how to fix it. Looks like gssapi.h is missing when running `cargo doc`

alexcrichton:
Documentation is built with `--all-features`, which enables `spnego` here which wasn't enabled before. This either needs to disable the feature on that builder or otherwise needs to install the necessary support to build with `spnego` on that builder.

kathampy:
Do I need to modify the Dockerfile to install libgssapi-krb5-2? I haven't worked with builders on GitHub before.

alexcrichton:
Documentation isn't built in a container, so you'll likely just need to `apt-get install` the right package

kathampy:
I don't understand where to do this in the source.

alexcrichton:
You'll probably need to edit `.github/workflows/main.yml` to update the failing builder to run a command to install the required dependencies

kathampy:
GSSAPI should work out of the box on Ubuntu. There's probably something wrong in the `curl-sys` build script for `libcurl`.

This auto-run CI on every commit seems grossly inefficient.

sagebind:
> This auto-run CI on every commit seems grossly inefficient.

How else would you have confidence whether a PR is mergeable or not? It's a very common pattern... 😉

kathampy:
> How else would you have confidence whether a PR is mergeable or not? It's a very common pattern... 😉

It seems to waste a lot of energy installing rust and doing a clean build every single time instead of reusing the previous setup. A clean run can probably be done on demand only when finally required.

kathampy:
Closing this PR as it may be better to use a custom `libcurl` in the path with additional features like `c-ares`.

sagebind:
I believe the correct package to install here is [`libkrb5-dev`](https://packages.ubuntu.com/bionic/libkrb5-dev).

I'm interested in this as well. @kathampy if you don't want to bother with this PR, I can make my own with installing `libkrb5-dev` for CI.

kathampy:
> I'm interested in this as well. @kathampy if you don't want to bother with this PR, I can make my own with installing `libkrb5-dev` for CI.

I'm using a custom `libcurl`, so I won't attempt to modify `curl-rust`.

PR #304, opened 2019-10-01
Enable HAVE_GETPEERNAME when building `curl`

alexcrichton:
Looks like this enables a few optimizations internally, such as reusing
connections! Let's be sure to enable this for our platforms, which
should all have this, so we can get those performance optimizations with
HTTP.

### DIFF
--- a/curl-sys/build.rs
+++ b/curl-sys/build.rs
@@ -187,6 +187,7 @@ fn main() {
         .file("curl/lib/warnless.c")
         .file("curl/lib/wildcard.c")
         .define("HAVE_GETADDRINFO", None)
+        .define("HAVE_GETPEERNAME", None)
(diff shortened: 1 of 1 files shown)


alexcrichton:
For some background on this, I've been debugging why a `curl` upgrade in Cargo recently caused issues where connections to crates.io were not being reused. Turns out [this block](https://github.com/curl/curl/blob/ac830139dadf4abd48232e3eb5ce0c8d06d19dcf/lib/connect.c#L722-L730) wasn't run, never setting `ip_addr_str`, causing [this to always be run](https://github.com/curl/curl/blob/ac830139dadf4abd48232e3eb5ce0c8d06d19dcf/lib/url.c#L1106-L1109).

PR #305, opened 2019-10-03
Fix Curl_now() timer resolution

sagebind:
Configure libcurl to use more accurate means of time measurement for use with various calculations, such as `CURLINFO_TOTAL_TIME`.


alexcrichton:
Oops, thanks!

PR #313, opened 2019-11-13
Re-export spnego feature from curl-sys

### LINKED ISSUE #300: `spnego` feature in `curl` crate
Shouldn't there be a `spnego` feature in the base `curl` crate rather than just in `curl-sys`? I now need to include `curl-sys` with `spnego` in my project.

sagebind:
Fixes #300.


alexcrichton:
👍

PR #319, opened 2020-01-09
Add an environment variable for requesting static linking with libcurl

dmitmel:
Based on the exact same feature of `libz-sys` implemented here:

https://github.com/rust-lang/libz-sys/blob/510d1b6a8935d7de974a0bfb4d9764bae28bb474/build.rs#L24-L25

This is necessary to enabled/disable static linking based on the target platform, since Cargo doesn't support target-specific features for dependencies.

Also, since `curl-sys` depends on `libz-sys`, I think it would be also useful to make a feature called `static-libz` which enables the `static` feature in `libz-sys`. I can implement this in another PR.

### DIFF
--- a/curl-sys/build.rs
+++ b/curl-sys/build.rs
@@ -9,6 +9,8 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 
 fn main() {
+    println!("cargo:rerun-if-env-changed=CURL_SYS_STATIC");
+
     let target = env::var("TARGET").unwrap();
     let windows = target.contains("windows");
 
@@ -19,8 +21,10 @@ fn main() {
         return println!("cargo:rustc-flags=-l curl");
     }
 
-    // If the static-curl feature is disabled, probe for a system-wide libcurl.
-    if !cfg!(feature = "static-curl") {
+    let want_static =
+        cfg!(feature = "static-curl") || env::var("CURL_SYS_STATIC").as_ref() == Ok("1");
+    // If static linking is disabled, probe for a system-wide libcurl.
+    if !want_static {
         // OSX and Haiku ships libcurl by default, so we just use that version
         // so long as it has the right features enabled.
         if target.contains("apple") || target.contains("haiku") {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! I feel like though there's already such a myriad of ways to link to libcurl that I'd prefer to not add yet another way to configure this. Rationalizing with all the existing ways feels like it's just unnecessarily complicating things :(

dmitmel:
The problem is basically what I described - there is no way right now to enable static linking based on the target platform. This is useful when building cross platform applications - you'd want to reuse system `libcurl` on \*NIXes (macOS, Linux etc) and link statically on Windows because Windows doesn't provide `libcurl`.

Also, IMO I'm not proposing a new way of linking, instead I'm proposing a new way of enabling static linking (which is an existing way), in other words effect of `CURL_SYS_STATIC` is equivalent to enabling the feature `static-curl`.

alexcrichton:
Yes I realize it's another way of doing things but that's sort of just it, it's a new way of doing things on top of the litany of "things" that are already there. There's already so many options of how curl can be configured, and I'm hesitant to just keep adding whatever someone thinks of as a new possible option. Eventually this sort of configuration will end up collapsing under it's own weight.

dmitmel:
Okay then, but please help me in this case. I'm currently developing an application which uses curl to make HTTP requests (unsurprisingly). Windows doesn't provide libcurl, so for Windows static linking is required. macOS has a new enough version of libcurl in its default installation, so I'd like to reuse that. Also, AFAIK most Linux distributions recommend dynamically linking to make use of libraries installed via a package manager. Now here is the question: how would you implement this sort of platform-dependent static linking on the application level if you were me?

alexcrichton:
I'm sorry for the delay on getting back to this. I'm also sorry but I don't have a ton of time right now to help out with this. I don't have the energy myself to keep maintaining such a long list of ways to link to libcurl in this crate, so I'm going to stick to what's here for now. I'm sorry if it doesn't solve your use case at this time, but I don't have a better solution :(

dmitmel:
Nevermind. I implemented this by doing the following:

https://github.com/CCDirectLink/ccloader-installer/blob/4ccf6dc29804967235ae386609ca01dc3367ce33/Cargo.toml#L45-L48

Issue #333, opened 2020-05-23
curl_global_init footgun


sagebind:
I've decided to open this issue here for discussion.

An [Isahc](https://github.com/sagebind/isahc) user brought to my attention this week a double-free error occurring in a relatively simple program (issue https://github.com/sagebind/isahc/issues/189), which I managed to simplify even further to a program like this:

```rust
fn main() {
    let threads = 8;
    let mut handles = Vec::new();

    for _ in 0..threads {
        handles.push(std::thread::spawn(move || {
            let mut easy = curl::easy::Easy::new();
            easy.url("http://localhost:8080").unwrap();
            easy.perform().unwrap();
        }));
    }

    for handle in handles  {
        handle.join().unwrap();
    }
}
```

Depending on which versions of curl is used and which SSL engine, the above program can randomly crash with a SIGABRT on a double-free error; in particular, this is with libcurl 7.54.0 and LibreSSL 2.6.5, both system defaults on macOS 10.14 (and likely other versions as well), so the chance of using these versions is pretty high on that platform.

Obtaining a backtrace when such a crash occurs reveals that the bad call of `free()` is coming from LibreSSL from inside a call to `curl_easy_cleanup()`, but it is not necessarily a bug. In our example program, `curl_global_init()` is getting called by `curl::init()`, which is lazily called once by `Easy::new()`. This is violating the rules of this function according to the [libcurl documentation](https://curl.haxx.se/libcurl/c/libcurl.html) (snipped the most relevant bits):

> The basic rule for constructing a program that uses libcurl is this: Call curl_global_init, with a _CURL_GLOBAL_ALL_ argument, immediately after the program starts, while it is still only one thread and before it uses libcurl at all. Call curl_global_cleanup immediately before the program exits, when the program is again only one thread and after its last use of libcurl.
>
> [...]
>
> It isn't actually required that the functions be called at the beginning and end of the program -- that's just usually the easiest way to do it. It _is_ required that the functions be called when no other thread in the program is running.
>
> These global constant functions are _not thread safe_, so you must not call them when any other thread in the program is running. It isn't good enough that no other thread is using libcurl at the time, because these functions internally call similar functions of other libraries, and those functions are similarly thread-unsafe. You can't generally know what these libraries are, or whether other threads are using them.

[...]

PR #334, opened 2020-05-24
Experimental use of constructor functions to call init

sagebind:
As mentioned in #333, this is a potentially helpful addition that ensures that curl is initialized on the main thread by using constructor functions that get called by the OS before the current program's `main` is called.

This has the advantage that, assuming you are on one of the supported platforms, `init()` will be called safely, correctly, and automatically without concerning the user about the gotchas.

This does have some disadvantages:

- Constructor functions are always invoked, which means that simply including curl can slow down the startup of a user's program even if curl is only conditionally used, or used later in program execution.
- On platforms without a constructor implementation, the user still needs to initialize curl on the main thread. All the common platforms are covered though, so maybe this is a niche scenario.

There's no additional runtime cost to this implementation except on platforms without a constructor, where we do an extra atomic swap that isn't necessary. This is probably fixable with additional conditional compilation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -84,8 +84,7 @@ mod panic;
 pub fn init() {
     /// An exported constructor function. On supported platforms, this will be
     /// invoked automatically before the program's `main` is called.
-    #[cfg_attr(any(target_os = "linux", target_os = "android"), link_section = ".init_array")]
-    #[cfg_attr(target_os = "freebsd", link_section = ".init_array")]
+    #[cfg_attr(any(target_os = "linux", target_os = "freebsd", target_os = "android"), link_section = ".init_array")]
     #[cfg_attr(target_os = "macos", link_section = "__DATA,__mod_init_func")]
     #[cfg_attr(target_os = "windows", link_section = ".CRT$XCU")]
     static INIT_CTOR: extern "C" fn() = init_inner;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -80,6 +80,7 @@ mod panic;
 ///
 /// It's not required to call this before the library is used, but it's
 /// recommended to do so as soon as the program starts.
+#[inline]
 pub fn init() {
     /// An exported constructor function. On supported platforms, this will be
     /// invoked automatically before the program's `main` is called.
@@ -93,8 +94,7 @@ pub fn init() {
     static INIT: Once = Once::new();
 
     // We invoke our init function through our static to ensure the symbol isn't
-    // optimized away due to a rustc bug:
-    // https://github.com/rust-lang/rust/issues/47384
+    // optimized away by a bug: https://github.com/rust-lang/rust/issues/47384
     INIT.call_once(|| INIT_CTOR());
 
     #[cfg_attr(any(target_os = "linux", target_os = "android"), link_section = ".text.startup")]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -64,7 +64,7 @@ extern crate schannel;
 
 use std::ffi::CStr;
 use std::str;
-use std::sync::Once;
+use std::sync::{atomic::{AtomicBool, Ordering}, Once};
 
 pub use error::{Error, FormError, MultiError, ShareError};
 mod error;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -80,32 +80,53 @@ mod panic;
...
 
-        // Note that we explicitly don't schedule a call to
-        // `curl_global_cleanup`. The documentation for that function says
-        //
-        // > You must not call it when any other thread in the program (i.e. a
-        // > thread sharing the same memory) is running. This doesn't just mean
-        // > no other thread that is using libcurl.
+    // We invoke our init function through our static to ensure the symbol isn't
+    // optimized away by a bug: https://github.com/rust-lang/rust/issues/47384
+    INIT.call_once(|| INIT_CTOR());
```
alexcrichton, line 97: Perhaps this `INIT `could move into `init_inner`? That way we wouldn't need a second atomic I think?

sagebind: I don't think so, because we are not allowed to call `Once::call_once` inside a constructor function, because `Once::call_once` depends on the Rust runtime. But the constructor function will run before the Rust runtime is initialized.

sagebind: I could test it on the various platforms, but `Once` relies on the `std::thread` machinery to be available.

An alternative would be to replace `Once` with an equivalent that does not rely on the standard library. Then we could use the same concurrency primitive in both the constructor and in the lazy init call.

alexcrichton: AFAIK there's no real "runtime" requirement and by definition this is happening on just the main thread so there's no contention, so it won't be needed unless this is manually called after the main function?

sagebind: > AFAIK there's no real "runtime" requirement

Basically, we have to _know_ that `Once` will work correctly when used before these calls are made: https://github.com/rust-lang/rust/blob/master/src/libstd/rt.rs#L34-L45

> and by definition this is happening on just the main thread so there's no contention, so it won't be needed unless this is manually called after the main function?

Right, but we still do the lazy init check in `Easy` and friends, so we have to make sure that the lazy init calls don't initialize a second time with something like a `Once` or an atomic. The way the code is right now, the constructor functions will set a boolean atomic if they ran since they can't inform the `Once` of that fact.

sagebind: OK, based on my reading of the [`Once` source](https://github.com/rust-lang/rust/blob/master/src/libstd/sync/once.rs), it should be OK to use since there's no possibility for contention. If there were, it would call `thread::current()` which to me looks like it would not work as expected if invoked before `rt.rs`. I'm just not sure if its OK to rely on the `Once` implementation details like that.

alexcrichton: Even if `thread::current()` is called, however, it shoudl be valid to do so. That's lazily initialized if it's not already present.

sagebind: Maybe this is getting in the weeds a bit, but wouldn't that cause this assert to fail? https://github.com/rust-lang/rust/blob/master/src/libstd/sys_common/thread_info.rs#L40 Since `lang_start_internal` calls `thread_info::set`, but the `ThreadInfo` for the main thread would have been already lazily created by the prior call to `thread::current()`?

sagebind: The following program:

```rust
#[used]
#[cfg_attr(target_os = "macos", link_section = "__DATA,__mod_init_func")]
static CTOR: extern "C" fn() = init;

extern "C" fn init() {
    // Yay side effects
    std::thread::current();
}

fn main() {
    println!("main");
}
```

unconditionally panics on start on macOS with the following output:

```
thread '<unnamed>' panicked at 'assertion failed: c.borrow().is_none()', src/libstd/sys_common/thread_info.rs:40:26
fatal runtime error: failed to initiate panic, error 5
fish: 'cargo run' terminated by signal SIGABRT (Abort)
```

alexcrichton: Hm I would classify that as a bug in the standard library actually. In any case that seems fine to avoid it for now while we wait for that to get fixed. Want to fixup CI here and then we can merge?

sagebind: Yeah, the CI failures are from ctest being broken on nightly now because the `asm!` macro was removed. I can look into that but I'm not sure where to begin.

alexcrichton: Ah I was also seeing rustfmt issues. The issue looks like it's the `extprim` crate, which I don't really have time to dig into where it's coming from right now.

I think ctest is largely unmaintained by gnzlbg now. I started a fork of it at https://github.com/alexcrichton/ctest/tree/use-maybe-unini and we can add more patches there to use here if necessary.

sagebind: Yeah I just opened a separate issue for the extprim problem: #337

sagebind:
I removed the second atomic so it should be just as efficient as it was before now. I also went ahead and updated the documentation.

alexcrichton:
👍

PR #336, opened 2020-05-26
Doc: add example with question mark operator

### LINKED ISSUE #335: Alternate error handling
I noticed the examples are using `unwrap` for error handling. I was interested
in other approaches, so I came up with this example:

~~~rust
use curl::easy::{Easy, WriteError};
use std::error::Error;
use std::io::{stdout, Write};
fn main() -> Result<(), Box<dyn Error>> {
   let mut o = Easy::new();
   o.url("http://speedtest.lax.hivelocity.net")?;
   o.write_function(|y|
      stdout().write(y).or(Err(WriteError::Pause))
   )?;
   o.perform()?;
   Ok(())
}
~~~

I wanted to post here in case its helpful to anyone else. I noticed some have
had confusion about what to do in this case:

<https://github.com/jeanguyomarch/mktcb/blob/9dda5763/src/download.rs#L62-L69>

nu8:
Fix #335

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,12 +6,12 @@ libcurl bindings for Rust
 
 ## Quick Start
 
+Print a web page onto stdout:
+
 ```rust
 use std::io::{stdout, Write};
-
 use curl::easy::Easy;
 
-// Print a web page onto stdout
 fn main() {
     let mut easy = Easy::new();
     easy.url("https://www.rust-lang.org/").unwrap();
@@ -25,10 +25,31 @@ fn main() {
 }
 ```
 
+Or using question mark operator:
+
+```rust
+use std::io::{stdout, Write};
+use curl::easy::{Easy, WriteError};
+use std::error::Error;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    let mut easy = Easy::new();
+    easy.url("https://www.rust-lang.org/")?;
+    easy.write_function(|data| {
+        stdout().write_all(data).or(Err(WriteError::Pause))
+    })?;
+    easy.perform()?;
+
+    println!("{}", easy.response_code()?);
+    Ok(())
+}
+```
+
+Capture output into a local `Vec`:
+
 ```rust
 use curl::easy::Easy;
 
-// Capture output into a local `Vec`.
 fn main() {
     let mut dst = Vec::new();
     let mut easy = Easy::new();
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! I think there's an issue on CI though?

sagebind:
It's common practice nowadays to only show examples using the `?` operator, since using `.unwrap()` is usually not what you want in real software. Perhaps we could simply replace the existing example?

sagebind:
@alexcrichton The nightly builds are borked as of recently because a dependency of ctest uses `asm!`, which is now a hard error being replaced with `llvm_asm!`.

muv1:
I studied Rust error handling recently, and I agree with @sagebind. This is probably a religious issue, but I think panic in general should be avoided, and the question operator seems like a good way to do that.

I can work on replacing the examples rather than adding, if its agreed thats the right thing to do.

src/lib.rs
```diff
@@ -30,6 +29,21 @@
...
+//! use std::io::{stdout, Write};
+//! use curl::easy::{Easy, WriteError};
+//! use std::error::Error;
+//!
+//! let mut easy = Easy::new();
+//! easy.url("https://www.rust-lang.org/")?;
+//! easy.write_function(|data| {
+//!     stdout().write_all(data).or(Err(WriteError::Pause))
+//! })?;
+//! easy.perform()?;
```
sagebind, line 44:
```suggestion
//! easy.write_function(|data| {
//!     stdout().write_all(data)
//!         .map(|_| data.len())
//!         .map_err(|_| WriteError::Pause)
//! })?;
//! easy.perform()?;
//! # Ok::<(), curl::Error>(())
```

README.md
```diff
@@ -25,10 +25,31 @@ fn main() {
...
+```rust
+use std::io::{stdout, Write};
+use curl::easy::{Easy, WriteError};
+use std::error::Error;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    let mut easy = Easy::new();
+    easy.url("https://www.rust-lang.org/")?;
+    easy.write_function(|data| {
+        stdout().write_all(data).or(Err(WriteError::Pause))
```
sagebind, line 39: This doesn't compile because `write_all` returns `()` on success, not `usize`. Try this instead:

```suggestion
        stdout().write_all(data)
            .map(|_| data.len())
            .map_err(|_| WriteError::Pause)
```

nu8:
Good idea. However I have put all the effort into this that I am willing to. No disrespect, but in general I dont like doing PRs, because often then end up just sitting there, even if no bugs.

PR #339, opened 2020-06-03
add support for ssl proxy

ritzk:
add support for ssl proxy


src/easy/handler.rs
```diff
@@ -881,6 +881,30 @@ impl<H> Easy2<H> {
...
+        let cainfo = CString::new(cainfo)?;
+        self.setopt_str(curl_sys::CURLOPT_PROXY_CAINFO, &cainfo)
+    }
+
+    /// Set private key for HTTPS proxy
+    ///
+    /// By default this value is not set and corresponds to `CURLOPT_PROXY_SSLCERT`.
+    pub fn proxy_sslkey(&mut self, cainfo: &str) -> Result<(), Error> {
+        let cainfo = CString::new(cainfo)?;
+        self.setopt_str(curl_sys::CURLOPT_PROXY_CAINFO, &cainfo)
```
alexcrichton, line 905: It looks like these all specify `CURLOPT_PROXY_CAINFO`? And they all have the same `cainfo: &str` argument?

alexcrichton:
Thanks! If these are relatively new you may need to add cases [here](https://github.com/alexcrichton/curl-rust/blob/e30fe223d52b7e5b44c681c7254c73a02e8d9843/systest/build.rs#L63-L97) to configure whether the constant values are checked against C or not

alexcrichton:
Thanks!

PR #347, opened 2020-07-22
Add unix_socket_path that exposes more Unix socket control

sagebind:
The existing `unix_socket` method has some limitations in its design:

- You cannot unset a previously set path.
- You cannot set non-UTF-8 paths.

This adds a second method named `unix_socket_path` that instead takes an `Option<impl AsRef<Path>>` which more closely aligns with what libcurl accepts for this option. We could consider soft-deprecating `unix_socket`, but I left it alone for now.


alexcrichton:
Seems reasonable to me!

PR #356, opened 2020-10-02
Fix some Clippy warnings

brenov:


alexcrichton:
Thanks!

src/easy/handler.rs
```diff
@@ -392,6 +392,7 @@ struct Inner<H> {
 unsafe impl<H: Send> Send for Inner<H> {}
 
 /// Possible proxy types that libcurl currently understands.
+#[non_exhaustive]
```
sagebind, line 395: I'm not sure what our current minimum supported Rust compiler version is (or maybe we should specify a MSRV) but this attribute requires Rust 1.40+.

alexcrichton: I personally go for the previous stable release, but if you'd like I'd be ok reverting this.

sagebind: I don't have a strong opinion, but I know that having a MSRV is advised for stable or popular crates. Since we don't have an MSRV right now this isn't a problem, but maybe we should consider adopting one.

I'm OK with leaving this for now. If we wanted to pick an older Rust version more agressively if/when we establish a MSRV, we can always change it, since `#[non_exhaustive]` is semantically backwards- and forwards-compatible with the `#[doc(hidden)] __Nonexhaustive` pattern.

PR #358, opened 2020-10-03
upgrade dev dependency mio from version 0.6 to 0.7

alexanderkjall:
I need to upgrade the version of mio packaged in Debian and curl-rust is also packaged there, so here is an attempt to upgrade the mio dependency to version 0.7.

I'm far from an expert on the mio library, so any feedback is welcome.


alexanderkjall:
Running the tests locally shows that this code does a timeout in about 1 out of 10 runs, sorry for not testing this better, It's not ready to be merged.

atouchet:
Mio 1.0 is now out.

PR #360, opened 2020-10-08
Add Max Age CURL Option

gsquire:
This would ideally be added to support the feature addition mentioned in https://github.com/sagebind/isahc/issues/93.

I found the constant listed [here](https://github.com/curl/curl/blob/9d954e49bce3706a9a2efb119ecd05767f0f2a9e/include/curl/curl.h#L1955).


sagebind:
Constants are validated against system libcurl which is dependent on versions. To add an exclusion configuration for a new constant, you can add a new match here:

https://github.com/alexcrichton/curl-rust/blob/badb4487f305e8ef6d15452ec2a9ab68a2c32d46/systest/build.rs#L63-L156

In this case, [`CURLOPT_MAXAGE_CONN`](https://curl.haxx.se/libcurl/c/CURLOPT_MAXAGE_CONN.html) was added in version 7.65.0.

sagebind:
Thanks for the PR! Can we also expose this in the Rust interface? Something like this:

```rust

impl Easy2 {
    pub fn maxage_conn(&mut self, maxage: Duration) -> Result<(), Error> {
        // todo
    }
}
```

curl-sys/lib.rs
```diff
@@ -582,6 +582,8 @@ pub const CURLOPT_PROXY_CAPATH: CURLoption = CURLOPTTYPE_OBJECTPOINT + 247;
 pub const CURLOPT_PROXY_SSLCERT: CURLoption = CURLOPTTYPE_OBJECTPOINT + 254;
 pub const CURLOPT_PROXY_SSLKEY: CURLoption = CURLOPTTYPE_OBJECTPOINT + 256;
 
+pub const CURLOPT_MAXAGE_CONN: CURLoption = 288;
```
sagebind, line 585: This should be `CURLOPTTYPE_LONG + 288`. Curl's headers can be a bit hard to grok because of the macros, but the `CURLOPT` macro actually adds an offset to the supplied number when it expands.

gsquire: Ah, thanks for sharing. I'll update that.

sagebind (approved):
Awesome, looks good to me. Thanks for contributing!

PR #363, opened 2020-10-29
Probe for OpenSSL certificates only once

### LINKED ISSUE #362: Creating an easy handle polls file system every time for OpenSSL
While doing some profiling of [Isahc](https://github.com/sagebind/isahc) to see if I could squeeze any more performance out of it, I noticed that creating new easy handles was spending a noticeable amount of time in `openssl_probe::probe`. Not an egregious amount of time, mind you, but enough to be noticeable. Digging deeper I saw that this function allocates a fair number of `PathBuf`s and makes `std::fs::metadata` calls to look for certificate files.

It is probably unlikely for these files to be changed at runtime, and if they do, I don't think most people would expect a running program to immediately pick up the change. Therefore I suggest that we lazily probe for certs just once, and re-use the results for subsequent easy handles created. Or at the very least, allow the user to opt-out of scanning for certs for every easy handle created.

sagebind:
Fixes #362.

### DIFF
--- a/src/easy/handler.rs
+++ b/src/easy/handler.rs
@@ -683,9 +683,7 @@ impl<H: Handler> Easy2<H> {
         INIT.call_once(|| unsafe {
             PROBE = Some(::openssl_probe::probe());
         });
-        let probe = unsafe {
-            PROBE.as_ref().unwrap()
-        };
+        let probe = unsafe { PROBE.as_ref().unwrap() };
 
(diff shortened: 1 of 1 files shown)


alexcrichton (approved):
Looks good! I think this needs a rustfmt though?

sagebind:
Missed that the first time, formatting is fixed now.

PR #365, opened 2020-11-30
Upgrade socket2 dependency

faern:
Upgrades to a version not making invalid assumptions about the memory layout of `std::net::SocketAddr`.

Helps unblock https://github.com/rust-lang/rust/pull/78802. See that PR for more details.


sagebind (approved):
Seems legit to me!

PR #373, opened 2021-02-08
Add environment variable to find GSSAPI

theawless:
This change allows us to set the location for GSSAPI.
This would be useful on systems where the library is not found at the usual location [`/usr`](https://github.com/curl/curl/blob/master/configure.ac#L1417) or if the client wants to set it manually.
I tried to make the behaviour similar to [`DEP_NGHTTP2_ROOT`](https://github.com/theawless/curl-rust/blob/master/curl-sys/build.rs#L218).

### DIFF
--- a/curl-sys/build.rs
+++ b/curl-sys/build.rs
@@ -344,6 +344,10 @@ fn main() {
                 .file("curl/lib/curl_gssapi.c")
                 .file("curl/lib/socks_gssapi.c")
                 .file("curl/lib/vauth/spnego_gssapi.c");
+            if let Some(path) = env::var_os("GSSAPI_ROOT") {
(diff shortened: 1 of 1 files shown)


alexcrichton:
FWIW the `DEP` variables are not meant to be set by users, they're set by Cargo because dependencies of this crate are named `nghttp2` and such (via the `links` key in the manifest). If this isn't going through Cargo for dependencies, could the env var be named differently?

theawless:
Would `GSSAPI_ROOT` be okay?

alexcrichton:
Seems fine by me!

theawless:
Updated the commits. Thank you for your time!.

Issue #341, opened 2020-06-13
Offer rustls as an optional TLS engine


sagebind:
It would be a large benefit to the Rust community if the curl crate could have an opt-in `rustls` feature that, when building the bundled libcurl, allows you to use [rustls](https://github.com/ctz/rustls) as a TLS backend.

This is possible to do by implementing a compatibility layer between libcurl and rustls by implementing libcurl's internal `vtls` interface -- rustls does not need to behave like OpenSSL or anything else, it only needs to perform the functions required by the `vtls` interface.

This compatibility layer could then be exposed as unmangled symbol names for libcurl to reference, which is legal since only one version of the curl crate can exist in a binary per Cargo's linking rules. To actually get libcurl to use the rustls backend, we could either:

1. Apply a small patch to the libcurl source code that adds the rustls-vtls symbols as an additional (and the selected) backend.
2. Have the rustls-vtls symbols impersonate a different, existing backend, and don't include the real backend's `.c` file when we compile.

I've done some investigation and experimentation into this concept myself and it seems feasible.

alexcrichton:
IIRC I think the `mesalink` feature was started to support this, but it may not be quite as buttery-smooth of an integration that would be great to have.

sagebind:
Yeah, MesaLink was looked to with promise, as it even made its way into libcurl officially. Unfortunately MesaLink does not support ALPN and cannot be used for HTTP/2 connections, and development seems to have been stalled with no communication for a while, unfortunately. We can still leave the door open for MesaLink as a viable solution if the developers come back.

sagebind:
Interesting update here: It looks like ISRG might be working on adding rustls as an official TLS engine to curl proper: <https://www.abetterinternet.org/post/memory-safe-curl/>. Of course, if that happens it should be relatively easy for us to adopt it.

sagebind:
Initial PR to add Rustls as an official opt-in backend upstream here: https://github.com/curl/curl/pull/6350. Once Rustls is released upstream, we can add a new crate feature here to opt-in to the Rustls backend. Unclear as of yet whether we will be able to pull in Rustls C bindings via Crates.io or we'll need to submodule, we'll have to wait and see.

sagebind:
This has now been merged in and will be exposed as the `rustls` crate feature in the next release. Since the curl integration with rustls isn't fully stable yet we won't offer any sort of stability or reliability guarantees with the `rustls` feature for now, but at least people will be able to start using it.

[...]

PR #374, opened 2021-02-10
Add experimental support for rustls

sagebind:
This adds experimental support for using rustls as a TLS backend for curl. This is made possible by the upstream rustls support recently added to curl via [rustls-ffi](https://github.com/rustls/rustls-ffi).

- [x] ~Upstream rustls was recently merged in as an experimental feature and likely is buggy. Is it OK for us to publish curl-rust with a `rustls` feature, knowing that it is experimental and may have serious bugs?~ Rustls support has greatly improved since being introduced, and docs are sufficient warning that implementation may not be fully complete yet.
- [x] ~Will the [crustls](https://github.com/abetterinternet/crustls) wrapper be willing to officially support our unusual use-case? Alternatively we can maintain a fork that is published to Crates.io as `curl-crustls` or equivalent.~ The `rustls-ffi` bindings package is now published to Crates.io.
- [x] ~Is it even correct to use `package.links` for crates that expose `no_mangle` symbols but don't actually link to anything external?~ Yes, this is appropriate.
- [x] Check rustls builds in CI
- [x] Update to a released curl version with support for rustls-ffi 0.8+

See #341.

### DIFF
--- a/examples/https.rs
+++ b/examples/https.rs
@@ -1,0 +1,23 @@
+//! Simple HTTPS GET
+//!
+//! This example is a Rust adaptation of the [C example of the same
+//! name](https://curl.se/libcurl/c/https.html).
+
+extern crate curl;
+
+use curl::easy::Easy;
+use std::io::{stdout, Write};
+
+fn main() -> Result<(), curl::Error> {
+    let mut curl = Easy::new();
+
+    curl.url("https://example.com/")?;
+    curl.write_function(|data| {
+        stdout().write_all(data).unwrap();
+        Ok(data.len())
+    })?;
+
+    curl.perform()?;
+
+    Ok(())
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,6 +58,10 @@ name = "atexit"
 harness = false
 
 [[example]]
+name = "https"
+path = "examples/https.rs"
+
+[[example]]
 name = "ssl_proxy"
 path = "examples/ssl_proxy.rs"
 required-features = ["ssl"]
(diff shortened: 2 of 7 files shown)


alexcrichton:
Nice! This seems reasonable to me to add!

I think it's fine to have off-by-default features which may be buggy if turned on, we can just make sure they're properly documented.

sagebind:
Status update: crustls now supports being built as an rlib dependency which means it can be used as a direct dependency to curl-sys. Crustls will not be published to Crates.io until its official home is found first, which is being discussed in https://github.com/ctz/rustls/issues/671. It is likely that it will be published to Crates.io under the `rustls-ffi` name. Once that is done then this PR will be mergeable for people to start experimenting with.

jsha:
rustls-ffi is now published: https://crates.io/crates/rustls-ffi. Thanks for your patience!

examples/https.rs
```diff
@@ -0,0 +1,23 @@
...
+fn main() -> Result<(), curl::Error> {
+    let mut curl = Easy::new();
+
+    curl.url("https://example.com/")?;
+    curl.write_function(|data| {
+        stdout().write_all(data).unwrap();
+        Ok(data.len())
+    })?;
+
+    curl.perform()?;
```
jsha, line 20: It would be nice to also verify the returned body bytes here, to guard against truncation or other corruption problems. Also, FWIW I like using httpbin.org for these sorts of public-Internet tests, in part because you can ask it to generate something with a known body. For instance, http://httpbin.org/#/Dynamic_data.

sagebind: I mostly added this as just a quick way of running an HTTPS request manually, it isn't meant to be considered a _test_. It is a direct translation of an upstream C example, which I can see being helpful and something I'll probably add more of in the future.

It would be nice to add an HTTPS check to the tests just for validation though.

sagebind:
Looks like curl 7.81.0 is slated to be released 2022-01-05: https://curl.se/dev/release-notes.html

sagebind:
Curl 7.81.0 was released today as scheduled, and since we've updated to that version on the `main` branch just hours ago, this feature is now ready to go and mergeable!

alexcrichton:
Nice!

jsha:
hooray! thanks for working on this!

PR #375, opened 2021-02-15
Haiku: rework build.rs logic for curl-sys

nielx:
Due to several issues, the Haiku build of rustc and cargo is cross-compiled
from a Linux host. In this official build, it is prefered to link to the
libcurl build that is being distributed as an official Haiku port. This
change adds the logic to link to the official libcurl build when
cross-compiling. The previous logic was flawed, in the sense that it would
rely on the hosts curl-config to determine whether http2 support was available.

When doing a native build, pkg_config can be used to determine whether
libcurl is available on the platform, thus removing the need for separate logic
when building curl-sys on Haiku itself.

The change was tested manually on Haiku and in a cross-compile from Linux.

### DIFF
--- a/curl-sys/build.rs
+++ b/curl-sys/build.rs
@@ -9,6 +9,7 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 
 fn main() {
+    let host = env::var("HOST").unwrap();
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #385, opened 2021-04-21
Add support for path_as_is curlopt

SpyMachine:
This adds support for [CURLOPT_PATH_AS_IS](https://curl.se/libcurl/c/CURLOPT_PATH_AS_IS.html) added in libcurl 7.42.0.

I saw that the bindings supported are for 7.24.0 and that another feature was added specifically for another option added in a later version of curl. Not sure how this would be best be handled but we can feature-ize this as well.


alexcrichton:
Thanks!

PR #389, opened 2021-05-02
Upgrade project to edition 2018

weihanglo:
Just update to edition 2018 before edition 2021 coming.

### DIFF
--- a/curl-sys/lib.rs
+++ b/curl-sys/lib.rs
@@ -1,17 +1,13 @@
 #![allow(bad_style)]
 #![doc(html_root_url = "https://docs.rs/curl-sys/0.3")]
 
-extern crate libc;
+// These `extern crate` are required for conditional linkages of curl.
 #[cfg(link_libnghttp2)]
 extern crate libnghttp2_sys;
 #[cfg(link_libz)]
 extern crate libz_sys;
-#[cfg(feature = "mesalink")]
-extern crate mesalink; // ensure lib is linked to
 #[cfg(link_openssl)]
 extern crate openssl_sys;
-#[cfg(windows)]
-extern crate winapi;
 
 use libc::c_ulong;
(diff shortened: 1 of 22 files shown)


curl-sys/lib.rs
```diff
@@ -1,17 +1,13 @@
...
 #[cfg(link_libnghttp2)]
 extern crate libnghttp2_sys;
 #[cfg(link_libz)]
 extern crate libz_sys;
-#[cfg(feature = "mesalink")]
-extern crate mesalink; // ensure lib is linked to
 #[cfg(link_openssl)]
 extern crate openssl_sys;
-#[cfg(windows)]
-extern crate winapi;
```
weihanglo, line 14: Here we preserve `extern crate` for openssl_sys, libnghttp2_sys and libz_sys for the sake of conditional linkage.

weihanglo:
Really weird. It seems that all CI tests has passed. https://github.com/alexcrichton/curl-rust/actions/runs/805297756

curl-sys/lib.rs
```diff
@@ -1,17 +1,13 @@
...
 #![doc(html_root_url = "https://docs.rs/curl-sys/0.3")]
 
-extern crate libc;
+// These `extern crate` are required for conditional linkages of curl.
 #[cfg(link_libnghttp2)]
 extern crate libnghttp2_sys;
 #[cfg(link_libz)]
 extern crate libz_sys;
-#[cfg(feature = "mesalink")]
-extern crate mesalink; // ensure lib is linked to
```
alexcrichton, line 10: I think this needs to remain for the linkage too?

alexcrichton:
Thanks for this!

curl-sys/lib.rs
weihanglo: Probably not? I thought it would be linked when feature "mesalink" is on. Anyway, updated at 5768b0b.

Issue #392, opened 2021-05-05
Enabling aws-sigv4 integration


theawless:
Hi, I wanted to enable the [aws-sigv4 integration](https://curl.se/libcurl/c/CURLOPT_AWS_SIGV4.html) that was introduced with [curl version 7.75](https://github.com/curl/curl/pull/5703). Multiple cloud providers use this scheme, but many don't have adequate rust clients (SDKs or signing libraries). I think it would be pretty nifty to have this feature supported out of the box in curl-rust (and hopefully in [isahc](https://github.com/sagebind/isahc)). Q1. What are your thoughts on this?

Here's my work so far: https://github.com/alexcrichton/curl-rust/compare/master...theawless:aws-sigv4
* added the `aws-sigv4` method in the `Easy`, `Easy2`, and `Auth` structs
* added aws-sigv4 related options in curl-sys lib
* removed `CURL_DISABLE_CRYPTO_AUTH` from curl-sys build
* added crypto-related files (digest, sha256, hmac, etc) to curl-sys build

I noticed that removing `CURL_DISABLE_CRYPTO_AUTH` automatically starts building NTLM which I didn't intend to. The already existing flag `CURL_DISABLE_NTLM` didn't seem to work. I noticed that curl got rid of that flag [here in this commit](https://github.com/curl/curl/pull/6809). I haven't worked with curl before but I think it might be impacting other dependencies as well (at least the [code search](https://github.com/search?q=CURL_DISABLE_NTLM&type=code) shows many instances). Q2: Do you think that this could be reverted and should I start a discussion at their end?

alexcrichton:
Seems reasonable to me to add this! If this requires fiddly bits as part of the build that's ok, the current build script is pretty specialized to what this crate needs as-is and if it needs more it's expected that we'd need to pull in more files and update things. I don't know ahead of time what changes are necessary to support this, but figuring it out on a PR seems fine by me.

theawless:
@alexcrichton Just one last question before I close this issue and continue the work on a pull request.

Since [the commit](https://github.com/curl/curl/pull/6809) `CURL_DISABLE_NTLM` is not working.
Do you happen to know a for sure way to disable NTLM while building curl?

alexcrichton:
Nah unfortunately I do not know.

PR #401, opened 2021-07-31
Add support for aws-sigv4 authentication

theawless:
Used these platforms:
* Windows10 + MSVC 14.29
* macOS Big Sur + Apple clang 12.0
* Linux 5.4 + GCC 7.3

I wasn't sure what feature combinations to build curl-rust with to test properly.
So, I wrote [this script](https://github.com/theawless/curl-rust/commit/4d4fca5c54772d47e6376045adf56dc2226fc38c) which runs all of them (1024 combinations).
Would it be useful to add it in the CI somehow?

For macOS and Linux, all builds work fine.
For Windows, some of them are breaking, but they had been breaking before my changes. 
`mesalink` + `spnego` doesn't build together on Windows. Is this expected?

Closes https://github.com/alexcrichton/curl-rust/issues/392

### DIFF
--- a/curl-sys/build.rs
+++ b/curl-sys/build.rs
@@ -109,7 +109,6 @@ fn main() {
         .include("curl/lib")
         .include("curl/include")
         .define("BUILDING_LIBCURL", None)
-        .define("CURL_DISABLE_CRYPTO_AUTH", None)
         .define("CURL_DISABLE_DICT", None)
         .define("CURL_DISABLE_GOPHER", None)
         .define("CURL_DISABLE_IMAP", None)
@@ -127,6 +126,7 @@ fn main() {
         .define("HAVE_ASSERT_H", None)
         .define("OS", "\"unknown\"") // TODO
         .define("HAVE_ZLIB_H", None)
+        .define("HAVE_LONGLONG", None)
         .define("HAVE_LIBZ", None)
         .define("HAVE_BOOL_T", None)
         .define("HAVE_STDBOOL_H", None)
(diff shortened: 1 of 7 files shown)


curl-sys/build.rs
```diff
@@ -287,8 +298,10 @@ fn main() {
             .define("USE_THREADS_WIN32", None)
             .define("HAVE_IOCTLSOCKET_FIONBIO", None)
             .define("USE_WINSOCK", None)
+            .file("curl/lib/bufref.c")
             .file("curl/lib/system_win32.c")
             .file("curl/lib/version_win32.c")
+            .file("curl/lib/vauth/digest_sspi.c")
```
theawless, line 304: The name suggests that `digest_sspi` is `USE_WINDOWS_SSPI` specific feature. Should this be under [line 256](https://github.com/alexcrichton/curl-rust/pull/401/files#diff-4b9a5dbfe7ad27e69c2b850cb2b18e53643d175862f843a5f2d1194df83609a5L256) instead?

The same goes for `spnego_sspi` on [line 295](https://github.com/alexcrichton/curl-rust/pull/401/files#diff-4b9a5dbfe7ad27e69c2b850cb2b18e53643d175862f843a5f2d1194df83609a5L295) which already existed.

alexcrichton:
Seems reasonable to me, thanks!

I think it's ok to keep CI as is (building all the sets of features isn't always the most useful and can take a lot of time), and for file order in the build script it's pretty haphazard and not super principled, so wherever is fine.

PR #405, opened 2021-09-15
Bump curl submodule to 7.79.0

alexcrichton:


ehuss:
@alexcrichton This caused a build failure on `dist-x86_64-illumos`: https://github.com/rust-lang/rust/pull/88956#issuecomment-920579457

It's a little late so I can't dig in much right now.  I assume it is missing `.file("curl/lib/bufref.c")` somewhere, but it is not immediately obvious to me why this would affect illumos and not other unix-like targets.

alexcrichton:
Hm let me try to dig in and see what's going on

Issue #403, opened 2021-09-07
Expose the curl_multi_poll & curl_multi_wakeup APIs


darkprokoba:
Currently, the curl crate uses internally the older curl_multi_wait API for classic poll(2)-based async clients.

It would be quite useful, if we could use the newer curl_multi_poll API instead (or let the user decide).

I have an application that does a lot of HTTP PUTs where the source that produces the data gets quite slow occasionally. When curl asks for data, but we don't have anything to send yet, my read_function returns Err(ReadError::Pause) and at some point in the future, I can resume the transfer with unpause_read(), however, there could be up to a second delay between the moment more data is available and the moment the transfer is unpaused. IIRC, curl_multi_poll/wakeup solve this problem precisely while having the extra benefit that curl_multi_wakeup can be invoked in any thread (nice!).

Are there any plans to expose these functions? If I attempted to do the work myself, what additional steps do I need to take in terms of documentation, unit tests, etc?

sagebind:
Thanks for filing an issue! Firstly I'd like to point out that the goal of the curl crate is to expose APIs that match the original curl APIs as closely as possible while enforcing complete safety. So swapping what curl function is used under the hood without notice seems out of the question. It is up to the user to decide how exactly they want to use curl. [`Multi::wait`](https://docs.rs/curl/0.4.*/curl/multi/struct.Multi.html#method.wait) is a pretty direct call to `curl_multi_wait`, so if you want to use `curl_multi_poll` then it would need to be exposed as a new method on the `Multi` struct.

If exposed, I would expect `curl_multi_poll` to be exposed as a `Multi::poll` method. Exposing `curl_multi_wakeup` is a lot tricker, since it has different thread safety requirements as the rest of the `curl_multi_*` methods. Based on [the docs](https://curl.se/libcurl/c/curl_multi_wakeup.html) and source code, it _seems_ that this can be called in a `Send` and `Sync` context, so maybe we would introduce a new `MultiWakeup` struct which can easily be acquired from a `Multi` struct that implements both `Send` and `Sync` and has a `wakeup(&self) -> Result<(), MultiError> ` method.

The hard part is controlling lifetime of such a handle -- if the `Mutli` struct is dropped we'll want to make sure to invalidate all `MultiWakeup` handles, either at runtime or programmatically, since we don't want to pass a dangling pointer.

[...]

PR #413, opened 2021-10-08
Exposed the curl_multi_poll libcurl API behind a new feature flag: po…

darkprokoba:
…ll_7_66_0, see #403.

See also: https://curl.se/libcurl/c/curl_multi_poll.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ with various Cargo features:
 - `static-ssl`: Use a bundled OpenSSL version and statically link to it. Only applies on platforms that use OpenSSL. Disabled by default.
 - `spnego`: Enable SPNEGO support. Disabled by default.
 - `upkeep_7_62_0`: Enable curl_easy_upkeep() support, introduced in curl 7.62.0. Disabled by default.
+- `poll_7_68_0`: Enable curl_multi_poll()/curl_multi_wakeup() support, requires curl 7.68.0 or later. Disabled by default.
 
 ## Version Support
 


src/multi.rs
```diff
@@ -31,6 +31,10 @@ use crate::{Error, MultiError};
 pub struct Multi {
     raw: *mut curl_sys::CURLM,
     data: Box<MultiData>,
+
+    // Add extra field so that all MultiWakers can track the dropped state of this Multi.
+    #[cfg(feature = "poll_7_66_0")]
+    dropped: std::sync::Arc<std::sync::Mutex<bool>>,
```
alexcrichton, line 37: I think the best way to structure this would, instead of an `Arc<Mutex<T>>` to do something like:

* Create a `RawMulti` wrapper which isn't `pub` which contains the actual `CURLM` pointer. 
* This `RawMulti` wrapper would have `unsafe impl Send/Sync`.
* The `RawMulti` wrapper would have a destructor
* The `Multi` struct would store `Arc<RawMulti>` plus the `data` field it has today.
* The `MultiWaker` struct would store `Weak<RawMulti>`

I think that would handle all the synchronization fiddly bits here and handle destruction pretty gracefully as well.

src/multi.rs
```diff
@@ -593,6 +607,74 @@ impl Multi {
         }
     }
 
+    fn _timeout_i32(timeout: Duration) -> i32 {
```
alexcrichton, line 610: It's ok to not prefix this method with an underscore since it's already private anyway.

src/multi.rs
```diff
@@ -593,6 +607,74 @@ impl Multi {
...
+    ///
+    /// let m = Multi::new();
+    ///
+    /// // Add some Easy handles...
+    ///
+    /// while m.perform().unwrap() > 0 {
+    ///     m.poll(&mut [], Duration::from_secs(1)).unwrap();
+    /// }
+    /// ```
+    #[cfg(feature = "poll_7_66_0")]
```
alexcrichton, line 655: Since this is a pretty special function, would it make sense to automatically enable this feature based on the detected curl version found at build time?

Otherwise though I think it's worth mentioning in the documentation that this is behind a feature, and I think we'll also want to update the docs.rs metadata for `Cargo.toml` to indicate that this feature should be enabled when documentation is generated.

alexcrichton:
Looks good to me! I think the only other thing is dealing with `poll_7_66_0`. If you'd like we can leave it, but I also think it's fine to remove it entirely. If it's used with a build of curl that doesn't have the function you'd get a link error, but that's somewhat reasonable to me so long as the documentation mentions it. Otherwise if it's not used you can still use this library with older versions.

PR #414, opened 2021-10-14
Enable the upkeep_7_62_0 and poll_7_68_0 features in CI

darkprokoba:
Enable the upkeep_7_62_0 and poll_7_68_0 features in CI whenever the curl_static feature is enabled.

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -5,19 +5,20 @@ set -ex
 cargo test --target $TARGET --no-run
 # First test with no extra protocols enabled.
 cargo test --target $TARGET --no-run --features static-curl
-# Then with all extra protocols enabled.
-cargo test --target $TARGET --no-run --features static-curl,protocol-ftp
+# Then with all extra protocols and features enabled.
+cargo test --target $TARGET --no-run --features static-curl,protocol-ftp,upkeep_7_62_0,poll_7_68_0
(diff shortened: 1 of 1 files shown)


darkprokoba:
@alexcrichton, if you're fine with @sagebind's proposal, i'll go ahead with it...

sagebind:
> I don't think the system you're talking about though @sagebind jives well with what we already have today. The majority of curl's new features go through options which means that it's always "available" you just get errors setting an unknown option if your version of curl is too old. In that sense we would need to apply this feature system to all of the option-setting methods as well, which seems pretty tedious.

Most of curl's new features _do_ go through options, and I think the current behavior of returning an error if not supported is perfectly fine and I think in the spirit of how curl operates, so I don't think we need to gate options by feature-versions (although if we were to adopt that it would make things more consistent overall). The problem is new function or struct definitions which will give you link-time errors or execution errors the developer can't recover from.

alexcrichton:
Oh I'm not saying we should give link-time errors, what I'm saying is that we can detect the version of `curl` found in the `curl-sys` crate and feed that into the `curl` crate which would then conditionally define functions like `wakeup` as always-return-an-error or call-the-actual-function depending on whether it's available or not.

Basically I'm pointing out how using future options should probably have the same story as using future methods where possible.

PR #416, opened 2021-10-16
Fix Clippy lints

winterqt:
This PR fixes various Clippy lints (and silences 5 of them), except for the one in `clang-sys`'s build script, which was fixed in #415.


src/easy/handle.rs
```diff
@@ -104,6 +104,7 @@ pub struct EasyData {
 unsafe impl Send for EasyData {}
 
 #[derive(Default)]
+#[allow(clippy::type_complexity)]
```
alexcrichton, line 107: I would personally prefer to not have lint annotation for clippy lints, so can this and others be removed?

PR #419, opened 2021-11-01
Update OpenSSL initialization check

alexcrichton:
As described in sfackler/rust-openssl#1548

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,13 +1,12 @@
 use std::env;
-use std::str::FromStr;
 
 fn main() {
     // OpenSSL >= 1.1.0 can be initialized concurrently and is initialized correctly by libcurl.
     // <= 1.0.2 need locking callbacks, which are provided by openssl_sys::init().
-    let use_openssl = match env::var("DEP_OPENSSL_VERSION") {
-        Ok(ver) => {
-            let ver = u32::from_str(&ver).unwrap();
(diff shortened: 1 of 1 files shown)


sfackler:
openssl_sys::init is a no-op on versions that don't need explicit initialization, so you can probably just remove the check here entirely and call it unconditionally.

ehuss (approved):
👍

sfackler:
There's also `DEP_OPENSSL_LIBRESSL_VERSION_NUMBER` if you care about supporting LibreSSL. Might be easier to invert the logic and only skip init if `DEP_OPENSSL_VERSION_NUMBER` is defined and >= 1.1.0.

alexcrichton:
I dug a bit more into this but I think this incantation is close to what this crate wants. While it'd probably be fine to call the openssl-sys init function unconditionally I think now with #418 and setting the probed env vars first it's best to do that conditionally when only OpenSSL is in used since that's all that needs that I think. Otherwise the goal of this check is already to skip init business if it's >= 1.1.0, and init should only be called for 1.0.1 and prior (or w/e the version was before 1.1.0)

alexcrichton:
I also think we can skip libressl since the purpose of this is to set the openssl-probe env vars and I don't think that's necessary for libressl, or at least no one's ever complained about that.

PR #420, opened 2021-11-06
Exposed the CURLOPT_CAINFO_BLOB libcurl option.

darkprokoba:
See also https://curl.se/libcurl/c/CURLOPT_CAINFO_BLOB.html


curl-sys/lib.rs
```diff
@@ -604,6 +604,7 @@ pub const CURLOPT_SSLKEY_BLOB: CURLoption = CURLOPTTYPE_BLOB + 292;
 pub const CURLOPT_PROXY_SSLCERT_BLOB: CURLoption = CURLOPTTYPE_BLOB + 293;
 pub const CURLOPT_PROXY_SSLKEY_BLOB: CURLoption = CURLOPTTYPE_BLOB + 294;
 pub const CURLOPT_ISSUERCERT_BLOB: CURLoption = CURLOPTTYPE_BLOB + 295;
+pub const CURLOPT_CAINFO_BLOB: CURLoption = CURLOPTTYPE_BLOB + 309;
```
sagebind, line 607: Minor nitpick: These constants are sorted in ascending order from the final additive offset and not by option type, so this constant should probably be at the end of the list. Generally in the same order as they are defined upstream in https://github.com/curl/curl/blob/master/include/curl/curl.h.

darkprokoba: Done.

Issue #421, opened 2021-11-06
curl::multi::Multi will make double free memory problem when some ssl error occurred


iiibui:
## Problem

**OS**: CentOS Linux release 7.3.1611 (Core)

Use curl::multi::Multi to fetch more than one https url will cause the program to crash when some ssl error occurred(rel [issue](https://github.com/rust-lang/cargo/issues/10034)):

```
[root@hugo-devm-b4pc3 easycurl]# cargo run
   Compiling easycurl v0.1.0 (/mnt/vdc1/home/apps/easycurl)
    Finished dev [unoptimized + debuginfo] target(s) in 0.58s
     Running `target/debug/easycurl`
result for easy with token 0: Some(Err(Error { description: "SSL peer certificate or SSH remote key was not OK", code: 60, extra: Some("SSL: certificate subject name 'Test' does not match target host name '127.0.0.1'") }))
*** Error in `target/debug/easycurl': double free or corruption (!prev): 0x00007f74f921f4e0 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x7c503)[0x7f74f6dec503]
/apps/svr/curl/lib/libcurl.so.4(+0x57d08)[0x7f74f7cc9d08]
/apps/svr/curl/lib/libcurl.so.4(+0x58c3c)[0x7f74f7ccac3c]
/apps/svr/curl/lib/libcurl.so.4(+0x1132c)[0x7f74f7c8332c]
/apps/svr/curl/lib/libcurl.so.4(curl_multi_cleanup+0xbf)[0x7f74f7cafbff]
target/debug/easycurl(_ZN63_$LT$curl..multi..RawMulti$u20$as$u20$core..ops..drop..Drop$GT$4drop17h76af921506dcf405E+0x12)[0x7f74f87a6a22]
target/debug/easycurl(_ZN4core3ptr42drop_in_place$LT$curl..multi..RawMulti$GT$17h3353ba6689bfbe2eE+0xb)[0x7f74f87a4f0b]
target/debug/easycurl(_ZN5alloc4sync12Arc$LT$T$GT$9drop_slow17h56cde95cf342e66dE+0x24)[0x7f74f87a2de4]
target/debug/easycurl(_ZN67_$LT$alloc..sync..Arc$LT$T$GT$$u20$as$u20$core..ops..drop..Drop$GT$4drop17hb73c8a2d4d00a004E+0x63)[0x7f74f87a2f43]
[...]

PR #423, opened 2021-11-11
Guarantee that curl_multi_remove_handle is called

sagebind:
Add a drop handler that guarantees that `curl_multi_remove_handle` is always called on easy handles added to a multi handle before the easy handle is dropped. Based on https://github.com/curl/curl/issues/7982, we should not be allowing users to drop a previously attached handle without first calling `curl_multi_remove_handle`. In curl versions between 7.77 and 7.80 this could cause a crash under certain circumstances. While this will likely be fixed in the next curl version, it is still recommended to not allow this, plus we must still handle this case for users who may not have updated to a patched libcurl.

I'm not totally happy with how this is implemented as it adds an `Arc::clone` call every time an easy handle is added to a multi handle, but I couldn't think of a better way of guaranteeing this behavior without breaking API changes.

I've verified that this does actually fix https://github.com/alexcrichton/curl-rust/issues/421 even without the upstream curl patch.

Fixes https://github.com/alexcrichton/curl-rust/issues/421.

### DIFF
--- a/src/multi.rs
+++ b/src/multi.rs
@@ -1027,9 +1027,9 @@ impl<H: fmt::Debug> fmt::Debug for Easy2Handle<H> {
     }
 }
 
-/// Guard that ensures that an easy handle is removed from a multi handle before
-/// the easy handle is dropped.
 impl DetachGuard {
+    /// Detach the referenced easy handle from its multi handle manually.
+    /// Subsequent calls to this method will have no effect.
     fn detach(&mut self) -> Result<(), MultiError> {
         if !self.easy.is_null() {
             unsafe {
--- a/src/multi.rs
+++ b/src/multi.rs
@@ -58,6 +58,8 @@ pub struct Message<'multi> {
 /// be used via `perform`. This handle is also used to remove the easy handle
 /// from the multi handle when desired.
 pub struct EasyHandle {
+    // Safety: This *must* be before `easy` as it must be dropped first.
+    guard: DetachGuard,
     easy: Easy,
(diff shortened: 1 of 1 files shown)


sagebind:
@iiibui Can you verify that using this branch fixes your crashes even with curl 7.80? It prevented the double-free errors in my tests.

iiibui:
> @iiibui Can you verify that using this branch fixes your crashes even with curl 7.80? It prevented the double-free errors in my tests.

This branch fixes my origin crashes with curl 7.80, but I found it has no effect with curl commit ```51c0ebcff2140c38ff389b4fcfb8216f5e9d198c```......

alexcrichton (approved):
Seems reasonable to me! In the grand scheme of things `Arc::clone` isn't too bad compared to HTTP so I'm not too too worried about the perf here myself.

sagebind:
@iiibui Hmm yeah, I still get the double-free against `51c0ebcff2140c38ff389b4fcfb8216f5e9d198c` as well, which first appears in curl 7.77.0. Seems like after this PR though we're doing things exactly as curl recommends and any other issues are likely on curl. Here's the behavior I see:

* curl 7.76.0: No crash regardless of this PR
* curl 7.77.0: Crashes, regardless of this PR
* curl 7.78.0-7.80.0: Crashes without this PR, works fine with this PR
* curl master now that https://github.com/curl/curl/issues/7982 is merged: No crash regardless of this PR

So the only released version with this issue still is 7.77.0, but that version had a known issue around that which was fixed in 7.78.0: https://github.com/curl/curl/issues/7236.

PR #426, opened 2021-11-17
Refactor init to avoid link bugs on macOS Monterey

### LINKED ISSUE #417: fails to link `curl::init()`'s call to `INIT_CTOR` on macOS 12.0 (Monterey)
### Summary

Only on macOS. May occur on macOS earlier than 12 using Xcode 13, I'm not sure. Underlying issue is https://github.com/rust-lang/rust/issues/90342

You get a linker error, there is a workaround for end users of the crate (see the rust issue; set the env var). I think curl-rust specifically can avoid this problem.

### Repro:

- rust 1.56 stable
- macOS 12.0.1 Monterey
-  `cargo test` in the curl-rust repo (currently at df64ee4).

This is using either system libcurl or with `--features static-curl`. Makes no difference.

```
$ cargo test

... very long cc -arch arm64 invocation
...
  = note: ld: reference to symbol (which has not been assigned an address) __ZN4curl4init9INIT_CTOR17h97cc33cf050cb462E in '__ZN4curl4init17ha644d831c2a57f65E' from /Users/cormac/git/tryout/libcurl-monterey/target/debug/deps/libcurl-0f9cbb7dde66dd88.rlib(curl-0f9cbb7dde66dd88.curl.0b6dcf6e-cgu.2.rcgu.o) for architecture arm64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

### Possible solutions, other than waiting for the rust issue

The relevant symbol is defined at:

https://github.com/alexcrichton/curl-rust/blob/df64ee4fcb1e3e2927d04e6bb6d8a1c0feedc844/src/lib.rs#L93-L101

I think the actual bug is due to the the workaround for https://github.com/rust-lang/rust/issues/47384 referenced further down, in which the `init` function calls `INIT_CTOR();`.

I think you can solve it by:

- Adding `#[used]` to `INIT_CTOR` (any MSRV issue? not sure)
- Moving `INIT_CTOR` outside the `init` function
- Change `pub fn init() { INIT_CTOR(); }` to `pub fn init() { init_inner(); }`

It works on macOS 12, at least. But I think it would be good enough, because:

- We're just trying to get rust to preserve `INIT_CTOR`, despite that rust-lang/rust issue 47384 where not using any code from the module means it won't get linked at all
- Does curl-rust have a problem with people using the library but managing not to use any code from it? If no code is used, do we need the pre-main initializer to run at all?
- I don't think so. It's not obvious to me that curl-rust is actually affected by the thing it has a workaround for.

sagebind:
Compilation currently fails for curl on macOS Monterey due to upstream rustc issue https://github.com/rust-lang/rust/issues/90342. To make this problem hurt users less, we can work around this by avoiding the specific issue that this bug causes.

To avoid the rustc issue, we cannot directly reference any symbol that is configured to be in a constructor linker section, which we were previously doing intentionally to work around a different rustc issue https://github.com/rust-lang/rust/issues/47384. We should be able to avoid both bugs by defining our constructor symbol as a public item in the root module, though not directly referencing it in other code. Since the root module is always used (`init` is called on-demand in key places in the code) it should not be removed by optimization.

Also add a quick unit test to make sure the constructor is still working for the platforms we have CI for.

Fixes #417.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -70,6 +70,9 @@ pub mod easy;
 pub mod multi;
 mod panic;
 
+#[cfg(test)]
+static INITIALIZED: std::sync::atomic::AtomicBool = std::sync::atomic::AtomicBool::new(false);
+
 /// Initializes the underlying libcurl library.
 ///
 /// The underlying libcurl library must be initialized before use, and must be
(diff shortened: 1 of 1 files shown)


alexcrichton:
👍

PR #428, opened 2021-11-26
Enable NTLM support

pbert519:
Support NTLM authentification

Not tested for windows


alexcrichton:
Since this is an optional feature could this be gated behind a Cargo feature?

pbert519:
I added the feature flag.

About the failed CI, i hit the scheduled brownout period, win2016 will be removed soon. See: https://github.com/actions/virtual-environments/issues/4312

PR #431, opened 2021-12-22
Added an example for the "mutli" api

reduf:
Hey,

here is a simple example to demonstrate how can one execute multiple requests concurrently in a non-blocking way.

I wasn't sure whether I should include urls in the source to test, but I've tested with the following commang:
```
> cargo r --example multi -- https://api.github.com/repos/rust-lang/rust https://api.github.com/repos/rust-lang/dskuyhfkuhdsfds "<invalid URL>"
```

You should observe a success (Code 200) on the valid repo, an explicit error (Code 404) on a non existing repo and no status in the case of the invalid url.

Cheers,


examples/multi.rs
```diff
@@ -0,0 +1,74 @@
...
+        std::thread::sleep(Duration::from_millis(16));
+    }
+
+    Ok(statuses)
+}
+
+fn main() -> Result<()> {
+    let urls: Vec<String> = env::args().skip(1).collect();
+
+    curl::init();
```
alexcrichton, line 65: I think that this can be skipped in examples?

examples/multi.rs
```diff
@@ -0,0 +1,74 @@
...
+        multi.perform()?;
+
+        multi.messages(|message| {
+            let token = message.token().expect("failed to get the token");
+            assert!(token <= statuses.len());
+            statuses[token] = parse_response(&mut handles[token], &message);
+            completed += 1;
+        });
+
+        std::thread::sleep(Duration::from_millis(16));
```
alexcrichton, line 56: I don't think that this `std::thread::sleep` is correct, all blocking should happen through the `Multi` interface itself.

alexcrichton (commented):
Thanks! Is there perhaps a `multi` example from curl itself we could lift? I think it'd be good to do something either with input going into requests or the output from the requests rather than simply dropping them on the floor.

examples/multi.rs
reduf: I noticed that it's not necessary under specified scenario described in the doc.

I ended up adding it, because of the following quote:
> ...it is strongly recommended that you call this function manually as soon as your program starts.

I don't have a big issue either way.

examples/multi.rs
reduf: The rational was that the curl multi interface is particularly adapted to combine in an application that already has an "update" that has to be called regularly. (e.g., every frames)

In this particular example it's indeed better to let the multi interface block as it will react to the io events if there is any.

sagebind:
> Is there perhaps a `multi` example from curl itself we could lift?

Curl has a few good C examples here: https://curl.se/libcurl/c/example.html. It would be nice to translate some of these into Rust and include them here.

examples/multi.rs
alexcrichton: I think we may want to tone down the recommendation there, but for small examples it shouldn't be necessary (or it's a bug in the `curl` crate)

sagebind: You missed the entire sentence in your quote (emphasis mine):

> This should be sufficient for most applications and scenarios, **but in any other case**, it is strongly recommended that you call this function manually as soon as your program starts.

It is strongly recommended that you call it manually only in the scenarios that are not already covered. Perhaps it could be reworded to be more clear, but basically, "If your program _doesn't_ match these requirements, then you should call it manually."

Since the example is single-threaded and isn't targeting uncommon platforms it is not necessary to call.

examples/multi.rs
alexcrichton: Sleeping is basically never the correct way to block, however, so this needs to be done by different means.

reduf:
> Curl has a few good C examples here: https://curl.se/libcurl/c/example.html. It would be nice to translate some of these into Rust and include them in our repo.

Indeed, I will adapt my example following one if theirs. [This one](https://curl.se/libcurl/c/10-at-a-time.html) seems to be in the same spirit.

examples/multi-dl.rs
```diff
@@ -0,0 +1,83 @@
...
+                        handle.get_ref().0.len()
+                    );
+                }
+                Err(error) => {
+                    println!("E: {} - <{}>", error, URLS[token]);
+                }
+            }
+        });
+
+        multi.wait(&mut [], Duration::from_secs(1))?;
```
alexcrichton, line 79: I think this `wait` only needs to happen if `still_alive` is true, right? Also I think that the duration here can be much longer because it's just a timeout which isn't being used for anything else here.

examples/multi-dl.rs
```diff
@@ -0,0 +1,83 @@
...
+    handle.set_token(token)?;
+    Ok(handle)
+}
+
+fn main() -> Result<()> {
+    let mut multi = Multi::new();
+    let mut handles: HashMap<_, _> = URLS
+        .iter()
+        .enumerate()
+        .map(|(token, url)| (token, download(&mut multi, token, url).unwrap()))
```
alexcrichton, line 40: Could this `.unwrap()` instead propagate the result?

alexcrichton (commented):
Seems reasonable to me! I've got one small comment but otherwise seems ok.

examples/multi-dl.rs
reduf: I'm not sure I understand correctly which part should unwrap?

Do you mean unwrapping in `download` such that the return value is an `Easy2Handle`?

alexcrichton (approved):

PR #434, opened 2022-02-02
Expose DNS-over-HTTP options

sagebind:
Expose options related to DNS-over-HTTP name resolution support. Includes a brief working example. Support for DoH is enabled in curl by default and is available as of 7.62.0.


alexcrichton (approved):
Nice!

PR #438, opened 2022-03-05
Update to curl 7.82.0 and remove MesaLink

sagebind:
Curl has removed support for MesaLink as of 7.82.0, so we can no longer build libcurl with MesaLink. But removing the feature from our crates would be a breaking change. Since this feature is not significantly used (I only found one public crate using it) I changed the feature to simply emit a warning and use the default TLS configuration.

Alternatively, we could remove the `mesalink` feature entirely and put a disclaimer in the release notes. I don't think it is worth bumping a major release of our crate for such an under-used feature, though I'd be open to alternate opinions.


alexcrichton (approved):
Seems reasonable to me! I think the build script may need updating though with the updated source structure

Issue #436, opened 2022-02-14
Expose CURLOPT_HTTP09_ALLOWED


jbangelo:
The [`CURLOPT_HTTP09_ALLOWED` option](https://curl.se/libcurl/c/CURLOPT_HTTP09_ALLOWED.html) tells curl to accept HTTP/0.9 responses, which can be a response just a body and no headers. This option has been disabled by default starting in v7.66 of curl, though some use cases still require it to be enabled.

sagebind:
This is now published in the [0.4.44 release](https://github.com/alexcrichton/curl-rust/releases/tag/0.4.44).

PR #447, opened 2022-05-04
Expose the CURLOPT_HTTP09_ALLOWED libcurl option

jbangelo:
This PR addresses my feature request in #436. It adds the ability to set the [`CURLOPT_HTTP09_ALLOWED`](https://curl.se/libcurl/c/CURLOPT_HTTP09_ALLOWED.html) option. Unfortunately there are still use cases for this option.

I took my best guess as to how to properly implement it. Please let me know if there's anything that needs to be changed, or even if exposing this option is acceptable.


src/easy/handle.rs
```diff
@@ -1280,6 +1280,11 @@ impl Easy {
         self.inner.pipewait(wait)
     }
 
+    /// Same as [`Easy2::http_09`](struct.Easy2.html#method.http_09)
+    pub fn http_09(&mut self, allow: bool) -> Result<(), Error> {
```
sagebind, line 1284: Nit: I might name this `http_09_allowed` which would more closely match the libcurl option name.

sagebind:
Thanks! The implementation looks correct to me, just some minor suggestions.

jbangelo:
@sagebind I've updated the method names as you suggested. Thanks for the quick feedback!

sagebind (approved):
Looks great!

PR #452, opened 2022-06-24
Forced lib usage for builds on windows

genusistimelord:
[Added in forced lib usage for vcpkg builds on windows](https://github.com/alexcrichton/curl-rust/commit/a8ae5076a8959a7d122c994e4f3c0454d739069b)

This will allow usage of pre-built curl libraries on windows. Works with vcpkg binaries.


genusistimelord:
@alexcrichton Any reason this should not be added to curl?

genusistimelord:
since no one cares.

sagebind:
This flag seems really specific and I'm not sure I understand its uses fully and the implications. Forcing the flag to do nothing except on Windows also seems like a strange behavior to me and not very composable.

For context, I'm always pretty wary about PRs that change how we build/link to curl, because we _already_ have tons of options around it and yet we haven't pleased everyone. People always have different ideas on how to link to curl, or to customize the build, or to use specific paths, etc, and unless the solution is very generic it is better for users to customize curl to their needs if they are that specific.

I suppose in this example we already have a `force-system-lib-on-osx` feature, so maybe this is acceptable. Though for naming consistency, perhaps `force-system-lib-on-windows` would be better.

PR #469, opened 2022-09-30
support ABSTRACT_UNIX_SOCKET

ktmlm:
this PR add supports for this feature: https://curl.se/libcurl/c/CURLOPT_ABSTRACT_UNIX_SOCKET.html


src/easy/handler.rs
```diff
@@ -807,6 +807,25 @@ impl<H> Easy2<H> {
...
+    /// Provides the ABSTRACT UNIX SOCKET which this handle will work with.
+    ///
+    /// The `addr` param can be any bytes.
+    ///
+    /// This function is an alternative to [`Easy2::unix_socket`] and [`Easy2::unix_socket_path`] that supports
+    /// ABSTRACT_UNIX_SOCKET(`man 7 unix` on Linux) address.
+    ///
+    /// By default this option is not set and corresponds to
+    /// [`CURLOPT_ABSTRACT_UNIX_SOCKET`](https://curl.haxx.se/libcurl/c/CURLOPT_ABSTRACT_UNIX_SOCKET.html).
+    pub fn abstract_unix_socket(&mut self, addr: &[u8]) -> Result<(), Error> {
```
sagebind, line 819: Make sure you expose the same option in `Easy` in `handle.rs` as well!

ktmlm: done

src/easy/handler.rs
```diff
@@ -807,6 +807,28 @@ impl<H> Easy2<H> {
...
+    /// ABSTRACT_UNIX_SOCKET(`man 7 unix` on Linux) address.
+    ///
+    /// By default this option is not set and corresponds to
+    /// [`CURLOPT_ABSTRACT_UNIX_SOCKET`](https://curl.haxx.se/libcurl/c/CURLOPT_ABSTRACT_UNIX_SOCKET.html).
+    ///
+    /// NOTE: this API can only be used on Linux OS.
+    #[cfg(target_os = "linux")]
+    pub fn abstract_unix_socket(&mut self, addr: &[u8]) -> Result<(), Error> {
+        unsafe {
+            self.cvt(curl_sys::curl_easy_setopt(
```
sagebind, line 824: `CURLOPT_ABSTRACT_UNIX_SOCKET` requires the parameter to be a null-terminated C string, so we should use `self.setopt_str` internally. As is, this is unsafe since we aren't enforcing `addr` to have a null-terminator.

sagebind (changes requested):

src/easy/handler.rs
ktmlm: yeh, the documentation indeed requires this rule, but non-null-terminated bytes can also work well in my practice; also, Linux's system-level interface does not have this requirement either.

ok, I will change the param to Cstr to meet the requirements of CURLOPT

ktmlm: done, please review again

sagebind: If it makes more sense, we can still accept `&[u8]` as the parameter type, but internally we should still use a `CString` to ensure memory safety. Technically, C strings don't really have any encoding requirements other than that it must end with a null byte, so this should be fine. We just want to make sure that the _caller_ is not responsible for adding the null terminator, because if they forget to do so it could lead to undefined behavior.

ktmlm: https://docs.rs/nix/0.25.0/src/nix/sys/socket/addr.rs.html#807-830

sagebind: Internally libcurl is going to call `strlen` and `strdup` on whatever gets passed in, which is why Very Much Bad Things™ could happen if we don't absolutely guarantee that there's exactly one null byte at the end of the slice:

- https://github.com/curl/curl/blob/10fbd8b4e3f83b967fd9ad9a41ab484c0e7e7ca3/lib/setopt.c#L2936-L2940
- https://github.com/curl/curl/blob/10fbd8b4e3f83b967fd9ad9a41ab484c0e7e7ca3/lib/setopt.c#L59-L76

Even if under the hood Linux doesn't care.

ktmlm: you are right, and I have pushed the required changes yesterday

src/easy/handle.rs
```diff
@@ -170,6 +170,14 @@ impl Easy {
         self.inner.unix_socket_path(path)
     }
 
+    /// Same as [`Easy2::abstract_unix_socket`](struct.Easy2.html#method.abstract_unix_socket)
+    ///
+    /// NOTE: this API can only be used on Linux OS.
+    #[cfg(target_os = "linux")]
+    pub fn abstract_unix_socket(&mut self, addr: &str) -> Result<(), Error> {
```
sagebind, line 177:
```suggestion
    pub fn abstract_unix_socket(&mut self, addr: &[u8]) -> Result<(), Error> {
```

src/easy/handler.rs
```diff
@@ -807,6 +807,21 @@ impl<H> Easy2<H> {
...
+    ///
+    /// This function is an alternative to [`Easy2::unix_socket`] and [`Easy2::unix_socket_path`] that supports
+    /// ABSTRACT_UNIX_SOCKET(`man 7 unix` on Linux) address.
+    ///
+    /// By default this option is not set and corresponds to
+    /// [`CURLOPT_ABSTRACT_UNIX_SOCKET`](https://curl.haxx.se/libcurl/c/CURLOPT_ABSTRACT_UNIX_SOCKET.html).
+    ///
+    /// NOTE: this API can only be used on Linux OS.
+    #[cfg(target_os = "linux")]
+    pub fn abstract_unix_socket(&mut self, addr: &str) -> Result<(), Error> {
```
sagebind, line 820:
```suggestion
    pub fn abstract_unix_socket(&mut self, addr: &[u8]) -> Result<(), Error> {
```

sagebind (commented):
This is what I mean that we can still use `&[u8]` since that is less restrictive and matches what is supported.

ktmlm:
good idea, done again

sagebind (approved):
Thanks!

PR #471, opened 2022-10-27
Update to 7.86.0

ehuss:
This updates to curl 7.86.0

This includes the fix that was causing issues with http2/tls 1.3/Windows 11 that caused the last release to be yanked. I have tested on Windows 11 and it seems to be working now.

Changelog: https://curl.se/changes.html#7_86_0

There are 4 CVE fixes in this release:

CVE-2022-42916: HSTS bypass via IDN
CVE-2022-42915: HTTP proxy double-free
CVE-2022-35260: .netrc parser out-of-bounds access
CVE-2022-32221: POST following PUT confusion

This release also adds support for websockets. However, I did not add any explicit support for that, though.

### DIFF
--- a/curl-sys/Cargo.toml
+++ b/curl-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "curl-sys"
-version = "0.4.57+curl-7.85.0"
+version = "0.4.58+curl-7.86.0"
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks!

PR #480, opened 2022-11-23
Make zlib optional

zengxs:
Add zlib feature for make it possible to disable zlib.


alexcrichton:
Thanks! Can you update CI to test this as well?

zengxs:
@alexcrichton 

I added unit tests for these cases: no-zlib, zlib, zlib-ng.
But the build of zlib-ng break ci, so I tried to introduce a step in `run.sh` to install cmake to solve this.

After hours for test and 10+ commits, CI is finally working. (It's such a pain to fight with github ci compile environment.)

Can you see if the current changes meet our needs?

ci/run.sh
```diff
@@ -2,6 +2,40 @@
 
 set -ex
 
+# install cmake. cmake required for zlib-ng build, and old cmake may not work,
+# so we install the latest one. this step only run for docker tests.
+if [ -f /.dockerenv ]; then
+    CMAKE_URL=https://github.com/Kitware/CMake/releases/download/v3.25.0/cmake-3.25.0-linux-x86_64.tar.gz
```
alexcrichton, line 8: This is only done in a docker environment so could this be part of the container build steps? Most of the curl/wget/apt-get stuff shouldn't be necessary since the container can just install one and run with that.

tests/zlib.rs
```diff
@@ -0,0 +1,48 @@
...
+    assert_eq!(curl::Version::get().feature_libz(), false);
+    assert!(curl::Version::get().libz_version().is_none());
+}
+
+#[cfg(all(
+    feature = "static-curl",
+    feature = "zlib",
+    not(feature = "zlib-ng-compat")
+))]
+#[test]
```
alexcrichton, line 17: Instead of having separate tests for this can this use `cfg!(...)` to avoid conditional compilation and always run the tests?

alexcrichton (commented):
Can this be added to the `default` set of features as well to preserve the historical behavior?

ci/run.sh
zengxs: There are 6 different docker images here, and they are based on centos:7 and ubuntu:16.04 respectively, I think if we do this step in the Dockerfile, there will be 6 copies of almost the same script code.

tests/zlib.rs
zengxs: ok let me change it

ci/run.sh
alexcrichton: You can add a `*.sh` script and then run the script in the container as well which should deduplicate this.

zengxs: Ok let me change it

zengxs:
@alexcrichton For historical compatibility, I set `static-curl` to enable zlib by default, and added a new feature `static-curl-no-zlib` to disable zlib.

ci/install-cmake.sh
```diff
@@ -0,0 +1,34 @@
...
+        apt-get update
+        apt-get install -y --no-install-recommends wget
+    elif command -v yum >/dev/null 2>&1; then  # centos
+        yum install -y wget
+    else
+        echo "No package manager found, cannot install wget to download cmake"
+        exit 1
+    fi
+    DL="wget -qO-"
+fi
```
alexcrichton, line 28: Sorry my main point was that I don't understand this complicated logic to figure out a `$DL` executable. Every docker file starts with a `sudo apt-get install ...` and I think `curl` can just be added there and this can all be removed inf favor of just using `curl`

Cargo.toml
```diff
@@ -45,7 +45,8 @@ mesalink = ["curl-sys/mesalink"] # MesaLink TLS backend
 http2 = ["curl-sys/http2"]
 spnego = ["curl-sys/spnego"]
 rustls = ["curl-sys/rustls"]
-static-curl = ["curl-sys/static-curl"]
+static-curl = ["curl-sys/static-curl", "curl-sys/zlib"]
+static-curl-no-zlib = ["curl-sys/static-curl"]
```
alexcrichton, line 49: I don't think that a new `static-curl-no-zlib` feature is necessary, could a new `zlib` feature be added which is added to `default`?

zengxs: Do you mean there should be a `zlib` feature here, which is included in the `default` set, but only takes effect when `static-curl` is turned on.
If we want to disable `zlib`, we need to set `default-features` to `false` like this:

```toml
[dependencies]
curl = { version = "*", default-features = false, features = ["static-curl"] }
```

sagebind: Technically moving a feature that used to always be available behind a feature under `default` is a breaking change. Any crate which was already setting `default-features = false` and expecting zlib to be included when doing will suddenly _not_ have zlib support.

Issue #484, opened 2023-01-09
Add the ability to use Pkg_config instead of Vcpkg when building with Mingw64  on Windows


lepapareil:
Hi @alexcrichton, and many thks for your work 😀.

I am having the same need as @ian-p-cooke had 4 years ago -> https://github.com/alexcrichton/curl-rust/issues/221.
I decided to use `mingw64` to build curl.dll on Windows instead of `vcpkg`. It works as expected following these steps:

- installing msys2
- installing base-devel and mingw-w64-x86_64-toolchain
- building curl 7.87 schannel from source for mingw64

Then I build my rust app with `crate-sys` dependency with these prerequisites:

- adding mingw64/bin dir to path
- rust toolchain stable-gnu

I expected `curl-sys` to link my `mingw64/curl-4.dll`, but it does not permit the use of mingw64's `pkg_config`. Instead, the crate uses embedded curl 7.86-DEV version as mentioned in documentation.

So, to achieve my goal, I had to update `curl-sys/build.rs` line `38` like this:

```
if windows {
  if try_vcpkg() {
    return;
    } else if try_pkg_config() {
        return;
      }
  } else if try_pkg_config() {
    return;
  }
}
```

With this update, `curl-sys` correctly links my rust app with my fresh `mingw64/curl-4.dll`.

Now that compiling curl with mingw64 is working easily, I think that we could allow `curl-sys` to use `pkg_config` instead of `vcpkg` according to the choice of devs.

So... can we reopen this issue please 😉 ?

theoparis:
I would also like to be able to use pkg-config on other platforms such as Linux to use the system libcurl, which I have built with aws-lc. This would also allow using curl with HTTP/3 support.

PR #486, opened 2023-02-10
Add the ability to use pkg_config on windows build

lepapareil:
#484

### DIFF
--- a/curl-sys/build.rs
+++ b/curl-sys/build.rs
@@ -39,6 +39,8 @@ fn main() {
         if windows {
             if try_vcpkg() {
                 return;
+            } else if try_pkg_config() {
+                return;
             }
         } else if try_pkg_config() {
             return;


lepapareil:
Hi @alexcrichton  and @sagebind can you review this change please ? Many Thks 😀.

alexcrichton:
Sorry this is one I don't personally have the time to invest in at this time. Changing build behavior of longstanding platforms often comes with breakage perhaps months down the line that I'm not in a position to manage at this time personally.

lepapareil:
I understand.

curl-sys/build.rs
```diff
@@ -39,6 +39,8 @@ fn main() {
         if windows {
             if try_vcpkg() {
                 return;
+            } else if try_pkg_config() {
```
micolous, line 42: I'd only run `try_pkg_config` when targeting non-MSVC targets.

micolous (approved):
[I made a similar change to `rust-openssl` recently](https://github.com/sfackler/rust-openssl/issues/1984), and this change looks reasonable to me with two suggestions:

1. I'd only attempt using `pkg-config` with non-MSVC targets.

2. Add a CI target which cross-compiles from Linux on a mingw toolchain, and running the tests in Wine.

   Unfortunately Ubuntu doesn't have a pre-packaged mingw `libcurl`, so you'd need to build `curl` from source _outside_ of this project, disable the auto-vendoring in `build.rs` (to ensure you're definitely using your "system libcurl"), and setup the pkg-config paths correctly for it.

With those changes, I think this PR should be good:

* when targeting Windows, `curl-rust` will continue to try using `vcpkg` first (so should have no effect if you use `vcpkg`); so this will only activate if `vcpkg-rs` can't find a matching package for the target.

* `pkg-config-rs` _by default_ won't allow cross-compiles (and potentially mixing host and target `.pc` files) unless you set up certain `pkg-config` environment variables for cross-compilation, or you set `PKG_CONFIG_ALLOW_CROSS=1` -- so it's effectively opt-in there.

* mingw has separate environments and prefixes for each build toolchain (eg: `mingw64`, `clangw64`, etc.) in an attempt to avoid cross-contamination.

The risk is if `static-curl` is disabled *and* `try_pkgconfig` somehow returned an invalid configuration (because of a misconfigured build environment), now it won't fall back to building a vendored `curl` from the git submodule.  Setting the `static-curl` feature would work around this.

curl-sys/build.rs
micolous: Actually, on second thought, this is fine; will explain in a comment on this PR.

micolous:
It looks like since I wrote my suggestion about not using `pkg-config` with MSVC, this has actually been fixed up: https://github.com/rust-lang/pkg-config-rs/pull/72, so it can be re-enabled (eg: https://github.com/rust-lang/libz-sys/pull/39)

PR #493, opened 2023-04-23
Allow to use static-ssl on windows-gnu targets

glandium:


glandium:
Unfortunately, it turns out I don't think that's possible to do because of https://github.com/rust-lang/cargo/issues/1197 : the default feature enables ssl, which enables openssl-sys, which is not actually wanted by default on windows-gnu. openssl-sys would only be wanted with static-ssl. I tried using different names (e.g. `foo = { package = "openssl-sys" }`), but cargo doesn't like that (`the crate ... depends on crate openssl-sys multiple times with different names`)

glandium:
Well, I think there would be a way to make this work... with an extra intermediate crate. ugh.

glandium:
Another option could be to enable both openssl and schannel backends on windows-gnu, with schannel as the default, but curl_global_sslset is not exposed.

PR #495, opened 2023-05-07
Use CURLOPT_TIMEOUT if timeout too big

LorenzoLeonardo:
i32 has a max of 2147483647, if the timeout duration is too big, it causes an overflow.

### DIFF
--- a/src/easy/handler.rs
+++ b/src/easy/handler.rs
@@ -1925,10 +1925,15 @@ impl<H> Easy2<H> {
     /// By default this option is not set and corresponds to
     /// `CURLOPT_TIMEOUT_MS`.
     pub fn timeout(&mut self, timeout: Duration) -> Result<(), Error> {
-        // TODO: checked arithmetic and casts
-        // TODO: use CURLOPT_TIMEOUT if the timeout is too great
-        let ms = timeout.as_secs() * 1000 + timeout.subsec_millis() as u64;
-        self.setopt_long(curl_sys::CURLOPT_TIMEOUT_MS, ms as c_long)
+        let ms = timeout.as_millis();
+        match c_long::try_from(ms) {
+            Ok(amt) => self.setopt_long(curl_sys::CURLOPT_TIMEOUT_MS, amt),
+            Err(_) => {
+                let amt = c_long::try_from(ms / 1000)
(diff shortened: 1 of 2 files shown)


LorenzoLeonardo:
Hello @alexcrichton  and @sagebind  kindly review this change. Thank you very much

src/easy/handler.rs
```diff
@@ -1924,10 +1924,13 @@ impl<H> Easy2<H> {
...
-        // TODO: use CURLOPT_TIMEOUT if the timeout is too great
-        let ms = timeout.as_secs() * 1000 + timeout.subsec_millis() as u64;
-        self.setopt_long(curl_sys::CURLOPT_TIMEOUT_MS, ms as c_long)
+        let mut ms = timeout.as_secs() * 1000 + timeout.subsec_millis() as u64;
+        if ms > std::i32::MAX as u64 {
+            ms = ms / 1000;
+            self.setopt_long(curl_sys::CURLOPT_TIMEOUT, ms as c_long)
+        } else {
+            self.setopt_long(curl_sys::CURLOPT_TIMEOUT_MS, ms as c_long)
+        }
```
alexcrichton, line 1933: Could this perhaps be modeled as:

```rust
match c_long::try_from(ms) {
    Ok(amt) => self.setopt_long(curl_sys::CURLOPT_TIMEOUT_MS, amt),
    Err(_) => {
        let amt = c_long::try_from(ms / 1000).unwrap();
        self.setopt_long(curl_sys::CURLOPT_TIMEOUT, amt)
    }
}
```

that avoids baking in knowledge about `i32` and `c_long`. Additionally it avoid confusingly storing a seconds-based value in a variable naed `ms`. Finally it asserts that the seconds conversion is indeed lossless.

alexcrichton:
Thanks! I left a few other follow up thoughts on https://github.com/alexcrichton/curl-rust/pull/497 to handle a few more edge cases here, and if those are applied here too I can merge. If you're up for it I'd also appreciate test which assert the error cases here too.

alexcrichton (approved):

LorenzoLeonardo:
It seems there are test error not related to the change.

LorenzoLeonardo:
it is ok now

alexcrichton:
Ah I think this needs a rebase but otherwise looks good!

PR #497, opened 2023-05-09
Use CURLOPT_CONNECTTIMEOUT if timeout too big

LorenzoLeonardo:
If the time that was converted into millisecond
is more than the capacity of c_long, it will
overflow, thus we need to convert it into seconds
and use CURLOPT_CONNECTTIMEOUT.

### DIFF
--- a/src/easy/handler.rs
+++ b/src/easy/handler.rs
@@ -1,4 +1,5 @@
 use std::cell::RefCell;
+use std::convert::TryFrom;
 use std::ffi::{CStr, CString};
 use std::fmt;
 use std::io::{self, SeekFrom, Write};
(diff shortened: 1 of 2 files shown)


src/easy/handler.rs
```diff
@@ -2049,7 +2050,14 @@ impl<H> Easy2<H> {
     /// `CURLOPT_CONNECTTIMEOUT_MS`.
     pub fn connect_timeout(&mut self, timeout: Duration) -> Result<(), Error> {
         let ms = timeout.as_secs() * 1000 + timeout.subsec_millis() as u64;
```
alexcrichton, line 2052: While not 100% related to this change, while you're here would you be up for using [`Duration::as_millis`](https://doc.rust-lang.org/stable/std/time/struct.Duration.html#method.as_millis) here instead? The fallible u128-to-`c_long` conversion below will automatically handle out-of-range values of `Duration` itself.

src/easy/handler.rs
```diff
@@ -2049,7 +2050,14 @@ impl<H> Easy2<H> {
...
-        self.setopt_long(curl_sys::CURLOPT_CONNECTTIMEOUT_MS, ms as c_long)
+
+        match c_long::try_from(ms) {
+            Ok(amt) => self.setopt_long(curl_sys::CURLOPT_CONNECTTIMEOUT_MS, amt),
+            Err(_) => {
+                let amt = c_long::try_from(ms / 1000).unwrap();
```
alexcrichton, line 2057: On second thought (after commenting on https://github.com/alexcrichton/curl-rust/pull/495), could this `.unwrap()` be translated to an error to be returned here instead? That way it's possible for callers to be notified that timeouts are simply too large.

LorenzoLeonardo: I tried to search the best error code to handle this for curl, but I cannot find that is related to overflow errors.

alexcrichton: Perhaps `CURLE_BAD_FUNCTION_ARGUMENT`?

LorenzoLeonardo: Please check, I added a Unit Test too

alexcrichton (approved):
I think tests may be failing though? (perhaps because curl is too old in the container?)

LorenzoLeonardo:
> I think tests may be failing though? (perhaps because curl is too old in the container?)

I will try to use i32:MAX instead of std::i32:MAX
https://doc.rust-lang.org/std/i32/constant.MAX.html

LorenzoLeonardo:
oh, that's weird, if I run the Unit Test locally this will succeed.

alexcrichton:
I believe the failure is because `c_ulong` is u64 on Linux so `u64::MAX` fits. I think you'll want to specify a larger amount than `u64::MAX` milliseconds to guarantee failure.

LorenzoLeonardo:
will try to use from_secs() to guarantee max value

LorenzoLeonardo:
It seems there are test error not related to the change.

LorenzoLeonardo:
it is ok now

PR #501, opened 2023-05-18
Update curl to 8.1.0

ecnelises:
curl 8.0.1 is vulnerable to CVE-2023-{28319..28322}, see https://curl.se/docs/security.html


alexcrichton:
Thanks!

PR #507, opened 2023-06-09
Remove mut to all getopt functions

LorenzoLeonardo:
There is no mutation involve when calling
getopt functions.


LorenzoLeonardo:
Hello @alexcrichton  please review this change. Mutable is not necessary for these functions.

alexcrichton (approved):
Thanks!

PR #509, opened 2023-06-27
 Allow running vcpkg on any Windows target

micolous:
`vcpkg-rs` doesn't support `-pc-windows-gnu` (mingw) _yet_, but its integration tests will try to build `curl-sys`, which currently prevents using `vcpkg` for non-`msvc` targets.

This change should have no effect for `curl-sys` right now, but [unblocks that work](https://github.com/mcgoo/vcpkg-rs/pull/52).

### DIFF
--- a/curl-sys/Cargo.toml
+++ b/curl-sys/Cargo.toml
@@ -35,12 +35,10 @@ openssl-sys = { version = "0.9.64", optional = true }
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3", features = ["winsock2", "ws2def"] }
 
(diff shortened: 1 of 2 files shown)


micolous:
@alexcrichton Any feedback on this? This is currently the last PR blocking https://github.com/mcgoo/vcpkg-rs/pull/52, because `curl-rust` is in its CI configuration.

alexcrichton:
Sorry for the delay, but I'm not confident myself signing off on this just yet. I would need to investigate the relevant packages and pieces involved here to ensure that everything works as I'd expect and I don't see any issues with this integration. I don't have the time to do this right now, though. @sagebind do you have thoughts on this though?

PR #532, opened 2023-10-11
Update to curl 8.4.0

ecnelises:
cURL 8.4.0 fixes two vulnerabilities:

- [CVE-2023-38545](https://curl.se/docs/CVE-2023-38545.html), high severity
- [CVE-2023-38546](https://curl.se/docs/CVE-2023-38546.html)

### DIFF
--- a/curl-sys/Cargo.toml
+++ b/curl-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "curl-sys"
-version = "0.4.67+curl-8.3.0"
+version = "0.4.68+curl-8.4.0"
 authors = ["Alex Crichton <alex@alexcrichton.com>"]
 links = "curl"
 build = "build.rs"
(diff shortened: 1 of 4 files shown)


ecnelises:
The macOS failure is reproducible locally, and previous commit is okay. Looking.

ecnelises:
https://github.com/curl/curl/commit/3aa3cc9b052353b12f57ae6b57ae5cd29c1060f1 breaks behavior of MIME multipart boundaries. I updated the tests and disabled them when using system curl.

ehuss:
I've been trying to investigate the `upload_lots` test hanging, but not getting too far. I bisected the change to https://github.com/curl/curl/pull/11939, which resulted in a report of tests hanging in https://github.com/curl/curl/issues/12033, but there were two fixes that supposedly fixed the hangs (which are in 8.4). 

I'm also wondering if the hanging is more related to mio, which is a very old version. Worringly, #358 tried to update, but ran into timeout problems, too.

If I can't make much progress soon, I think we should just disable the test for now to get the CVE fixes out. However, I don't have a grasp if that will make things worse if the hangs are legitimately a problem.

ehuss:
I pushed a fix for the timeout. I also pushed a slightly different approach for the multipart boundary size change which should work with whatever curl version is in use.

I'm going to go ahead and merge to get the CVE fixes out.

ecnelises:
Thanks!

PR #535, opened 2023-10-19
add example to use Easy2 in downloading files

LorenzoLeonardo:
Adding examples how to use Easy2 in downlaoding files


LorenzoLeonardo:
Hi @alexcrichton  and @sagebind  please add this example in using Easy2 on how to use to download files or return response as a body. This will be helpful for the community in using Easy2.

PR #539, opened 2023-11-26
Add new `CURLINFO_` constants and `fn response_http_version()`

### LINKED ISSUE #538: HTTP response version
Hi! I'm writing a little bit of [`http` interop code](https://crates.io/crates/http) in [a crate](https://github.com/ramosbugs/oauth2-rs/issues/236), and running into the need of the _HTTP response version_. I seem to only be able to _set_ a preferred HTTP version with this `Easy` portion of the crate.

I'm not super familiar with `curl`, but it seems [the bindings](https://curl.se/libcurl/c/curl_easy_getinfo.html) have `CURLINFO_HTTP_VERSION` for this, but I have no clue how new it is nor does it appear to be in the bindings already. Are the `curl-sys` bindings autogenerated or written by hand?

---

Note that we could separately discuss whether this `curl` crate should have built-in (behind an optional crate feature) `http` interop support, now that they've had their `1.0.0` release?

MarijnS95:
Fixes #538

Use `CURLINFO_HTTP_VERSION` (a new constant for the `curl-sys` crate) to determine the HTTP version used in the last/latest connection.


src/easy/handler.rs
```diff
@@ -3128,6 +3128,20 @@ impl<H> Easy2<H> {
...
+    /// Corresponds to `CURLINFO_HTTP_VERSION` and may return an error if the
+    /// option isn't supported.
+    pub fn response_http_version(&self) -> Result<HttpVersion, Error> {
+        self.getopt_long(curl_sys::CURLINFO_HTTP_VERSION)
+            .map(|c| match c as i32 {
+                curl_sys::CURL_HTTP_VERSION_1_0 => HttpVersion::V10,
+                curl_sys::CURL_HTTP_VERSION_1_1 => HttpVersion::V11,
+                curl_sys::CURL_HTTP_VERSION_2_0 => HttpVersion::V2,
+                curl_sys::CURL_HTTP_VERSION_3 => HttpVersion::V3,
+                _ => unreachable!("TODO"),
```
MarijnS95, line 3142: According to https://curl.se/libcurl/c/CURLINFO_HTTP_VERSION.html this only returns a select few codes, but should we map them all "just in case"? It can also return `0` in case the version cannot be determined, should we map that "none" to `Any` or wrap this in an `Option`? And should we wrap other faulty codes into a `Result` of sorts or panic on invalid library return values?

Sorry, still a bit new to this crate and `curl`, while adding this change on a tangent of a tangent :)

src/easy/handler.rs
```diff
@@ -3128,6 +3128,20 @@ impl<H> Easy2<H> {
             Ok(list::from_raw(list))
         }
     }
+    /// Get the last http version number
+    ///
+    /// Corresponds to `CURLINFO_HTTP_VERSION` and may return an error if the
+    /// option isn't supported.
+    pub fn response_http_version(&self) -> Result<HttpVersion, Error> {
```
sagebind, line 3135: This function ideally would be called `http_version` to match the curl constant, but perhaps instead should be called `get_http_version`. We want the names to match the underlying curl names as much as possible, even if it is less understandable.

src/easy/handler.rs
sagebind: We should assume that curl _could_ return anything here, and the last thing we want to do is panic when encountering an unknown result. Curl is often dynamically linked, meaning we don't have control over which version of libcurl is actually used. If old Rust code using old `curl-rust` code starts to run against a new libcurl version due to a system upgrade that starts returning a hypothetical "HTTP 4", then we:

1. For sure should not panic on any return value
2. If we can, return the underlying `i32` even though we don't understand it, allowing the caller to attempt to understand it themselves if they want to

One possible way of doing this would be to wrap the return value in a new enum that maybe looks something like this:

```rust
pub enum HttpVersionInfo {
    HttpVersion(HttpVersion),
    Unknown(i32),
}
```

MarijnS95: That conflicts a bit with the use of a `Result` type. I'd rather wrap the existing `Result<HttpVersion, CurlError>` with a `enum HttpVersionSpecificError { CurlError(CurlError), UnknownValue(i32) }` of sorts :)

MarijnS95: I ended up using your suggestion as `Unknown` isn't immediately invalid.

In another crate we've had discussions about a user hypothetically matching `Unknown(0x1337)`, and the crate being bumped with a semver-compatible release (or breaking release, doesn't really matter here) that maps the new `SuperDuperHttpVersion = 0x1337` value in this `#[non_exhaustive]` `enum`. The user will suddenly run into their mandatory `_ => {}` catch-all arm again. _Not_ giving them access to the value entices contributions and requests upstream to match the new value, and have less ambiguity inbetween.

src/easy/handler.rs
MarijnS95: Note that `get_` prefixes are discouraged in Rust, but let's just use it here.

src/easy/handler.rs
```diff
@@ -428,7 +428,7 @@ pub enum IpResolve {
 
 /// Possible values to pass to the `http_version` method.
 #[non_exhaustive]
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum HttpVersion {
```
MarijnS95, line 432: Curious why simple enums have to be compared by `match`/`if let` and are not allowed to be used in simple `==` by deriving these quite-standard `Eq` traits. Intentional or oversight?

MarijnS95:
Looks like the CI isn't too happy about the new constants added to `curl-sys`, will have to find a way to update "something" here.

PR #552, opened 2024-02-18
Un-pin nightly in CI.

jsirois:
Releases of garando_syntax 0.1.1 and ctest2 0.4.8 fix the issues that
required the pin.


alexcrichton (approved):
Thanks!

PR #556, opened 2024-03-26
use vcpkg only from `*-msvc` targets

ognevny:
for other targets (`*-gnu` and `*-gnullvm`) pkg-config is usually used

### DIFF
--- a/curl-sys/build.rs
+++ b/curl-sys/build.rs
@@ -7,6 +7,7 @@ fn main() {
     println!("cargo:rerun-if-changed=curl");
     let target = env::var("TARGET").unwrap();
     let windows = target.contains("windows");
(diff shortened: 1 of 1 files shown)


ognevny:
I've rebased branch and simplified patch a bit
@alexcrichton can this be merged? I see other -sys crates use vcpkg for MSVC only

alexcrichton:
I'm sorry but I don't personally have the energy to shepherd this. Changes like this unfortunately have a pretty high likelihood of causing breakage for users. That breakage typically requires follow-ups from the maintainership side here which I don't have a ton of time for nowadays. If someone else wants to champion this though and field requests though that's ok too! (they'll need to merge though)

PR #562, opened 2024-06-25
Update curl to 8.8.0

### LINKED ISSUE #559: abort_read test fails on Debian sid, probablly due to curl
Recently the abort_read test started failing on our CI in Debian sid. I do not know for sure what triggered this, but the prime suspect is the upgrade from curl 8.5 to curl 8.7 was the trigger for this regression.

I was able to reproduce this with a git checkout from this and the command "cargo test". libcurl4-openssl-dev was installed.

---- abort_read stdout ----
thread '<unnamed>' panicked at 'unexpected header: "" (remaining headers {"\r\n", "PUT / HTTP/1.1\r\n", "Host: 127.0.0.1:33559\r\n", "Content-Length: 2\r\n", "Accept: */*\r\n"})', tests/server/mod.rs:70:21
thread 'abort_read' panicked at 'res failed with Any { .. }', tests/server/mod.rs:218:13

ehuss:
A routine update.

Changelog: https://curl.se/changes.html#8_8_0

Fixes #559

### DIFF
--- a/curl-sys/Cargo.toml
+++ b/curl-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "curl-sys"
-version = "0.4.72+curl-8.6.0"
+version = "0.4.73+curl-8.8.0"
 authors = ["Alex Crichton <alex@alexcrichton.com>"]
 links = "curl"
 build = "build.rs"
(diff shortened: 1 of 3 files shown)


alexcrichton (approved):
Thanks!

Issue #568, opened 2024-08-14
Unable to connect to an IPv6 link local address


M-Tarasov:
A connection to link local address requires the scope ID to be set (e.g. 4 in https://[fe80::abcd:1234:abcd:1234%4]) in the `struct sockaddr_in6`. However, the symbol `HAVE_SOCKADDR_IN6_SIN6_SCOPE_ID` is not defined in the `curl-sys` crate, and the `scope_id` field is always set to 0.

See `cf-socket.c` -> `socket_open(..)`:
```c
#if defined(USE_IPV6) && defined(HAVE_SOCKADDR_IN6_SIN6_SCOPE_ID)
  if(data->conn->scope_id && (addr->family == AF_INET6)) {
    struct sockaddr_in6 * const sa6 = (void *)&addr->sa_addr;
    sa6->sin6_scope_id = data->conn->scope_id;
  }
#endif
```

M-Tarasov:
Added pull request https://github.com/alexcrichton/curl-rust/pull/569

PR #569, opened 2024-08-14
Enable support of  scope ID for IPv6 link local addresses

M-Tarasov:
Try to resolve issue https://github.com/alexcrichton/curl-rust/issues/568

### DIFF
--- a/curl-sys/build.rs
+++ b/curl-sys/build.rs
@@ -122,6 +122,7 @@ fn main() {
         .define("CURL_DISABLE_TFTP", None)
         .define("CURL_STATICLIB", None)
         .define("ENABLE_IPV6", None)
+        .define("HAVE_SOCKADDR_IN6_SIN6_SCOPE_ID", None)
         .define("HAVE_ASSERT_H", None)
         .define("OS", "\"unknown\"") // TODO
         .define("HAVE_ZLIB_H", None)


alexcrichton:
Thanks! Is it possible to add a test for this?

M-Tarasov:
> Thanks! Is it possible to add a test for this?

Unfortunately, I have no idea how it could be tested with unit tests: from `rust` point of view, it just on more define in build script for a C library, from the other side, the effect of this define could be observed deep in the `libcurl` library itself. It is possible, to test it in an integration test within specific environment, but it would require much effort to set up. I could just say, that this change is working for my project.

alexcrichton:
Does connecting with ipv6 not work at all without this? Or could a dummy server be set up with a different link-local ID and then curl tries to connect to it and it would fail if this isn't supported?

M-Tarasov:
> Does connecting with ipv6 not work at all without this? Or could a dummy server be set up with a different link-local ID and then curl tries to connect to it and it would fail if this isn't supported?

The issue is related to the IPv6 link-local addresses (all addresses within `fe80::/10` range) only. 

Just one example: let suppose you have a local web server running with `python3 -m http.server -b "::" 8123`. Your `eth0` interface has a link local IPv6 address `fe80::c4:d164:150d:282c`. Then you make standard curl calls like this:

```bash
curl http://[fe80::c4:d164:150d:282c%eth0]:8123  # works fine, with scope id eth0
curl http://[fe80::c4:d164:150d:282c]:8123       # will not work without scope id
```
Without the flag (`HAVE_SOCKADDR_IN6_SIN6_SCOPE_ID`) enabled, both variants would fail.

In the `libcurl` itself, the `HAVE_SOCKADDR_IN6_SIN6_SCOPE_ID` flag is defined whenever the system's `struct sockaddr_in6` has the field `sin6_scope_id` (no idea, where it could be missing, maybe some very old systems). See curl's original `CMakeLists.txt`: https://github.com/curl/curl/blob/curl-8_9_0/CMakeLists.txt#L291

PR #577, opened 2024-09-20
Use the Cargo.toml version of the check-cfg squelching

### LINKED ISSUE #576: error: unsupported output in build script of `curl-sys v0.4.76+curl-8.10.1`: `cargo::rustc-check-cfg=cfg(libcurl_vendored,link_libnghttp2,link_libz,link_openssl,)`
#575 accidentally increased the (undocumented) MSRV to 1.80, and older rust versions fail to compile.

glandium:
It doesn't make older version of cargo complain.

https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html#check-cfg-in-lintsrust-table

Fixes #576


alexcrichton:
Thanks! Mind adding the MSRV you're interested in to `.github/actions/main.yml` as well?

glandium:
My own MSRV is probably much more recent than what other people might be relying on, but 1.80 is really too recent. It would be great to figure out the real practical MSRV and decide what the real supported version should be.

alexcrichton:
I've got no preference myself, I mostly figure that if you don't want this to regress again it'd be good to test in CI. No guarantee has been given up to now, but now's better late than never!

Basically I think this should be tested in CI. As to exactly what version is tested, iunno what was the practical MSRV before so I'm fine if you just want to fill in what you're interested in. If someone else is so interested they can send a PR to reduce it further.

glandium:
This is actually more tricky than in looks: only versions of cargo >= 1.73.0 and < 1.80.0 complain about rustc-check-cfg output from build.rs. Versions older than cargo 1.73.0 still build curl-sys just fine... I'm not super convinced it's worth testing a version between 1.73.0 and 1.80 just for check-cfg now it's there in a way supported by these versions.

glandium:
The practical MSRV looks like it is 1.63.0, FWIW (from socket2 and cc).

glandium:
Could you merge this and do a release?

alexcrichton:
Can you add a CI entry for the MSRV you're interested in?

glandium:
See https://github.com/alexcrichton/curl-rust/pull/577#issuecomment-2364727389

alexcrichton:
I still don't really understand the consequences of that comment? Why would that mean that more tests can't be added to CI? Is the crate as-is broken on only that range of rustc versions? Does your desired MSRV not fall in that range?

glandium:
> Is the crate as-is broken on only that range of rustc versions? 

Yes.

> Does your desired MSRV not fall in that range?

Yes, but that's kind of incidental. And once this is fixed, there isn't really a reason it would break (except if the MSRV bump abruptly, but that would be better served with a test with 1.63, which doesn't anything for this issue (it works).

alexcrichton:
Can you add testing for your MSRV then? At the very least something needs to be added here if you'd like a guarantee it doesn't regress. If other older verisons are broken then others can send PRs to fix those if they're interested.

glandium:
How would you expect it to regress? Someone explicitly reverting the change?

alexcrichton:
I think it's a good idea to have CI regardless of if it's likely to break or not. It serves as a good documentation reminder for future contributors and additionally doubles-up about being extra-sure that mistakes aren't made.

alexcrichton:
I've merged and published https://github.com/alexcrichton/curl-rust/pull/581 and I've double-checked it works for 1.63. If you're interested I think it'd be reasonable to add to CI here (perhaps as a second entry since as you were saying it behaves differently than 73-80).

PR #578, opened 2024-09-20
Install cmake and nasm in the docker image cross compiling for *-windows-gnu

glandium:


glandium:
Looks like it's not working like it's supposed to.

Issue #586, opened 2024-11-16
Missing function `Easy::into_inner`


jscheid:
I can move (in the sense of transferring ownership) a handler into an Easy (via `Easy::new`), and the Easy into a Multi (via `Multi::add`).

 I can later move the Easy out of the Multi (via `Multi::remove`). But, I can't move the handler out of the Easy, I can only get a reference via `get_ref` or `get_mut`. It would be useful to be able to do that. I believe the function would be called `into_inner` by convention, like in `RefCell::into_inner`.

This applies to Easy2 as well. Would you accept a PR for adding the function in both places or is there a reason it doesn't exist?

sagebind:
This sounds like a useful addition to me, go for it!

PR #587, opened 2024-11-17
Add function into_inner to Easy and Easy2 (#586)

jscheid:


src/multi.rs
```diff
@@ -982,6 +982,11 @@ impl<H> Easy2Handle<H> {
         self.easy.get_mut()
     }
 
+    /// Consumes the Easy2 handle, returning the underlying handler.
+    pub fn into_inner(self) -> H {
+        self.easy.into_inner()
+    }
+
```
jscheid, line 989: Maybe this should return an Easy instead? It wouldn't be symmetric with `new` however, and not sure if there's much value in it.

src/easy/handler.rs
```diff
@@ -379,8 +379,12 @@ pub struct Easy2<H> {
     inner: Box<Inner<H>>,
 }
 
+/// A thin wrapper around a low-level Easy handle that performs
+/// cleanup when dropped.
+struct CURLPtr(*mut curl_sys::CURL);
```
jscheid, line 384: This is necessary because `Inner` can't be `Drop` anymore due to [E0509](https://doc.rust-lang.org/error_codes/E0509.html), and because the following workarounds are unavailable:
- We can't use `std::mem::take` because H isn't `Default`.
- We don't want to use an `Option` because it adds overhead.

src/easy/handler.rs
```diff
@@ -3698,6 +3723,20 @@ fn double_seconds_to_duration_sub_second2() {
     assert_eq!(dur.subsec_nanos(), 500_000_000);
 }
 
+impl CURLPtr {
+    fn as_ptr(&self) -> *mut curl_sys::CURL {
+        self.0
+    }
+}
```
jscheid, line 3730: Since the struct is package-private, perhaps this is too much ceremony and we should just reach into it with `.0` everywhere. I don't think there's a perf difference, mostly a matter of taste. Happy to remove this.

jscheid: Or we could `impl Deref` - maybe that would be nicer?

src/multi.rs
```diff
@@ -982,6 +982,11 @@ impl<H> Easy2Handle<H> {
         self.easy.get_mut()
     }
 
+    /// Consumes the Easy2 handle, returning the underlying handler.
+    pub fn into_inner(self) -> H {
+        self.easy.into_inner()
```
sagebind, line 987: Will this cause drops to be performed in the incorrect order? It seems like `self.easy.inner.handle` will be dropped first, triggering `curl_easy_cleanup`, and _then_ at the end of the function, `self.guard` will be dropped, triggering `curl_multi_remove_handle`. Whereas they need to happen in the opposite order.

To fix this, I believe calling `self.guard.detach()?` first before extracting the inner value should be sufficient.

jscheid: I'm trying to build a failing test case for this but without much luck:

Adding the Easy to the Multi moves ownership of the Easy into the Multi (since `Multi::add` takes a `mut Easy`). This means that `into_inner` can't be called by client code anymore after adding, since `into_inner` takes `self` and therefore requires ownership.

To regain ownership, the client code first has to call `Multi::remove`, meaning that the detach guard never comes into play.

I could be missing something?

sagebind: `Multi::add` takes ownership of `Easy2`, but then passes ownership to the `Easy2Handle`, which then owns the `Easy2`. Example:

```rust
use curl::{easy::{Easy2, Handler}, multi::Multi};

struct DummyHandler;

impl Handler for DummyHandler {}

fn main() {
    let easy = Easy2::new(DummyHandler);
    let multi = Multi::new();
    let handle = multi.add2(easy).unwrap();
    let handler = handle.into_inner();
}
```

And definition of `Easy2Handle::into_inner` annotated:

```rust
    pub fn into_inner(self) -> H {
        // took ownership of `self`
        self.easy.into_inner() // takes ownership of `self.easy`, and then drops it
        // `self` is dropped here, including `self.guard`
    }
```

And `Easy2::into_inner`:

```rust
    pub fn into_inner(self) -> H {
        // took ownership of `self`
        self.inner.handler
        // `self` is dropped here, including `self.inner.handle`
    }
```

If you add a `println!` where `curl_multi_remove_handle` and `curl_easy_cleanup` are called, you'll see in my example that `curl_easy_cleanup` is called first, and then `curl_multi_remove_handle`.

jscheid: Ah, sorry - I didn't mean to add `into_inner` to `Easy2Handle` in the first place. I've pushed a fixup commit to remove it, does it look better now?

PR #608, opened 2025-06-06
Bump curl/curl-sys versions

alexcrichton:
One day I might set up auto-publishing but alas it is not yet such a day.


ehuss (approved):
Thanks! I meant to bump this in https://github.com/alexcrichton/curl-rust/pull/606, but somehow messed it up.

PR #615, opened 2025-07-05
Rewrite test to remove outdated dependencies

Expyron:
The goal is mainly to remove the dev-dependency on a very old version of mio.


alexcrichton:
Ah yes the purpose is not just `assign`, but also a real event loop. Sleeping for the timeout duration is not valid because the purpose of an event loop is to wait for I/O *or* sleep for the timeout, whereas here it's considering all I/O always ready and also sleeping for the timeout.

I understand mio 0.6 is old, but this is also just a dev-dep in a project somewhere on the internet. Can you elaborate more on the rationale for updating the dependency? Is there more motivation beyond updating it?

alexcrichton:
If it's possible to use tokio yeah that'd be great, I'm happy to drop the mio dep but I don't want to lose what's being tested.

PR #620, opened 2025-08-12
Update curl to 8.15.0

ehuss:
This updates curl to 8.15.0.

Changelog: https://curl.se/ch/8.15.0.html
Blog: https://daniel.haxx.se/blog/2025/07/16/curl-8-15-0/

A fairly major change here is that support for SecureTransport on macOS has been dropped. I decided to switch it over to OpenSSL. I don't know how well this is going to work. Apple's version of curl uses LibreSSL, with some minor changes. The main concern is how well it will find certificate stores, since IIRC those can be stored in the Keychain. The basic tests of accessing regular websites seem to pass for me.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -63,7 +63,7 @@ jobs:
           - build: macos
             os: macos-latest
             rust: stable
-            target: x86_64-apple-darwin
+            target: aarch64-apple-darwin
           - build: win32
             os: windows-latest
             rust: stable


ehuss:
I cherry-picked https://github.com/alexcrichton/curl-rust/pull/618 to see if that will get CI to pass.

alexcrichton:
This seems reasonable enough to me, thanks! I suspect inevitably someone will open an issue about this but there's only so much that can be done with upstream changing.

Issue #400, opened 2021-07-22
Necessary function commented out


lputnam2000:
https://github.com/alexcrichton/curl-rust/blob/f56f767cc88e60963c222d74c39ad27264bc6f57/src/easy/handler.rs#L1382

PR #626, opened 2025-10-15
Implements proxy_headers method to add headers only to the proxy.

jcamiel:
This PR re-enables the `proxy_headers` method (see <https://curl.se/libcurl/c/CURLOPT_PROXYHEADER.html> available since curl 7.37.0) 

It is basically the same code as `http_headers`. 

The rational for this PR is to be used in [Hurl](https://hurl.dev), a cli HTTP engine powered by `curl-rust`. Hurl exposes various curl commands, basically delegating everything to `curl-rust/libcurl` and adding some syntax to test HTTP responses. One issue <https://github.com/Orange-OpenSource/hurl/issues/3944> and one draft PR <https://github.com/Orange-OpenSource/hurl/pull/4482> needs to use this new method exposed by `curl-rust`.

To test it, I've used this sample:

```rust
use curl::easy::{Easy, List};

#[test]
fn test_proxy_header() {
    let mut easy = Easy::new();
    easy.url("https://google.com").unwrap();
    easy.ssl_verify_peer(false).unwrap();

    let mut headers = List::new();
    headers.append("Proxy-Only: foo").unwrap();
    easy.proxy_headers(headers).unwrap();

    let mut headers = List::new();
    headers.append("All: bar").unwrap();
    easy.http_headers(headers).unwrap();

    easy.proxy("http://127.0.0.1:8080").unwrap();
    easy.verbose(true).unwrap();
    easy.write_function(|data| Ok(data.len())).unwrap();
    easy.perform().unwrap();
}
```

And with mitmproxy:

```shell
$ mitmproxy
```

We can see that only "All: bar" header is sent to the target :

mitmproxy


jcamiel:
Solve also this issue #400

jcamiel:
Hi all, would it be possible to have any kind of feedbacks on this PR, either positive or negative? It can be closed without explanation if not rekevant, thanks!!

ehuss (approved):
Seems good, thanks!

PR #643, opened 2026-03-28
fix(sys): link to `iphlpapi` unconditionally

weihanglo:
Upstream has done that so we follow:

1. curl/curl#17413 preloads `iphlpapi` conditioned on `HAVE_IF_NAMETOINDEX`
2. curl/curl@b17ef873 `HAVE_IF_NAMETOINDEX` is defined unconditionally

See rust-lang/rust#154482 which [failed without](https://github.com/rust-lang/rust/pull/154482#issuecomment-4148090768) and [succeeded with](https://github.com/rust-lang/rust/pull/154482#issuecomment-4148609723) link to `iphlpai` dll.

This also bumps curl-sys to 0.4.87 to help unblock rust-lang/rust#154482, (of course, if curl-rust maintainers can help release 🙏🏾)

### DIFF
--- a/curl-sys/Cargo.toml
+++ b/curl-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "curl-sys"
-version = "0.4.86+curl-8.19.0"
+version = "0.4.87+curl-8.19.0"
 authors = ["Alex Crichton <alex@alexcrichton.com>"]
 links = "curl"
 build = "build.rs"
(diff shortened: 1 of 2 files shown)


weihanglo:
https://github.com/weihanglo/cargo/pull/89 is now failing when raw-dylib is in use. 

I still need to double check whether bootstrap builds Cargo with raw-dylib though

ehuss (approved):
Ok, thanks! It's a little strange, but I think I gather this is due to behavior in the `windows-sys` and `windows-link` crates, and those crates no longer including everything with the `windows_raw_dylib` cfg.

bjorn3:
https://github.com/alexcrichton/curl-rust/pull/648 fixes raw-dylib support.

PR #646, opened 2026-04-01
Update MSRV to 1.82

ehuss:
libnghttp2-sys has been updated to use new syntax (`unsafe extern`) that is not available in 1.80.


alexcrichton (approved):
Er sorry about that, should probably track this better downstream...

PR #648, opened 2026-04-23
Allow using windows-sys 0.61

bjorn3:
This version no longer requires using big import libraries and instead uses raw-dylib.

### DIFF
--- a/curl-sys/build.rs
+++ b/curl-sys/build.rs
@@ -459,6 +459,8 @@ fn main() {
         println!("cargo:rustc-link-lib=ws2_32");
         println!("cargo:rustc-link-lib=crypt32");
         println!("cargo:rustc-link-lib=iphlpapi");
+        println!("cargo:rustc-link-lib=advapi32");
+        println!("cargo:rustc-link-lib=secur32");
     }
 
     // Illumos/Solaris requires explicit linking with libnsl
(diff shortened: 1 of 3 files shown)


bjorn3:
The windows CI failures look to be because curl itself depends on the import library for advapi32 and nothing is including it anymore. It previously got included through windows-targets. Added an explicit instruction to link against it.

alexcrichton:
@bjorn3 do you know what could cause the failure https://github.com/alexcrichton/curl-rust/actions/runs/25217899867/job/73942557471 on `main`? That was also when coupled with https://github.com/alexcrichton/curl-rust/pull/649, which I'm not sure if that changed things.

curl-sys/build.rs
```diff
@@ -459,6 +459,8 @@ fn main() {
         println!("cargo:rustc-link-lib=ws2_32");
         println!("cargo:rustc-link-lib=crypt32");
         println!("cargo:rustc-link-lib=iphlpapi");
+        println!("cargo:rustc-link-lib=advapi32");
+        println!("cargo:rustc-link-lib=secur32");
```
bjorn3, line 463:
```suggestion
        println!("cargo:rustc-link-lib=secur32");
        println!("cargo:rustc-link-lib=bcrypt");
```

probably fixes that issue.

alexcrichton: Worked perfectly, thanks!

PR #649, opened 2026-05-01
Update curl to 8.20.0

ehuss:
Changelog: https://curl.se/ch/8.20.0.html
Release blog: https://daniel.haxx.se/blog/2026/04/29/curl-8-20-0/

This fixes 8 CVEs.

Note that the next release will likely remove NTLM, so we'll need to figure out how to deal with that when that happens.


alexcrichton:
I attempted to combine this with https://github.com/alexcrichton/curl-rust/pull/648 to get that out as well, but that failed [on main](https://github.com/alexcrichton/curl-rust/actions/runs/25217899867/job/73942557471) which I'm pinging about [here](https://github.com/alexcrichton/curl-rust/pull/648#issuecomment-4359801348).

@ehuss is this something you think is worth more urgently getting out though due to the CVEs fixed? If so I can revert the windows-sys update and go ahead and publish.

ehuss:
No rush, though looks like he's identified the issue.

None of the CVEs look concerning to me.

PR #661, opened 2026-08-15
Add HTTP/3 support through the already stable ngtcp2 libcurl back-end

Cupertinum:
Adds an `http3` feature to curl-sys that builds/links libcurl against the ngtcp2 and nghttp3 backends.

libcurl no longer marks HTTP/3 via ngtcp2/nghttp3 as "EXPERIMENTAL" (<https://curl.se/docs/http3.html>), unlike its quiche implementation, so this seemed like a good time to add it.

I used the ngtcp2 and nghttp3 bindings from the trusted [http3 crate](https://github.com/0x676e67/http3), which is maintained by people that work on hyperium.


alexcrichton (commented):
Thanks for the PR, but I'm not prepared to pick up new major features like this in this crate. I do not have the bandwidth for picking up maintainership of this feature myself.

